PR #1, opened 2022-08-29
port to PyQt6

trel:
I was having trouble building/installing PyQt5 on Apple Silicon.

Turns out, a few things moved in PyQt6.  But this builds, and appears to work as well as before I touched anything.

PR #10, opened 2022-09-04
a few more things

trel:
builds atop #1


chStaiger:
I think everything apart from "prevent exploding when browsing collections" is now present in the current working and testing branch *pyqt6*?!

trel:
yes, seems good - will close this.   thanks!

Issue #9, opened 2022-09-04
iBridges should identify itself so that ips can be more helpful


trel:
`ips` currently shows `python-irodsclient`, the default populated value when a python-irodsclient-based application connects to the server.

chStaiger:
Fixed in branch *pyqt6*

PR #14, opened 2022-09-13
[9] add application_name, for ips monitoring

trel:
There is probably a better place to declare the application name only once.   I had to declare it twice in this PR.

I then use it in the Connection to iRODS.  This makes the application name appear in the `ips` output for the iRODS Zone.

I believe the work done in f4afed010b9e35ae50eba025beecb8d75d90f008 via `setproctitle` only sets the name for the local operating system.


trel:
part of #9

PR #18, opened 2022-10-10
Adding owner to ACL tab

chStaiger:


chStaiger:
Changes also mailmen.py to only a selection of features. Hence: reject

PR #20, opened 2022-10-18
Added deployment pipeline

wurDevTim:
Build the deployment pipeline and ran initial tests on windows. 
I visited all the menus, and everything looks good now, both when running from code and as executable. 
Please give it a try on ubuntu as well. 
For as far as I can see it should be fine, but I cannot check to be sure.


wurDevTim:
Heey Trel, 

It's definitely on the wish list for the new research data management team. 
But the first priority at WUR is to get iRODS running in production and verify if we want to work with iRODS or move to azure.
So currently we are only implementing to go/no-go criteria.

trel:
Of course - thanks!

chStaiger:
Tested on Mac and linux. Both with python 3.10.
For linux we might have to include the python library Pillow (dependency for elabjournal)

PR #22, opened 2022-10-26
Pyqt6

chStaiger:
- Port to PyQt6
- Add pyinstaller
- Add logging window to uploads
- New buttons (parent, refresh) in Browser


chStaiger (commented):
Looks all good

PR #62, opened 2023-03-04
checking if the irods session is valid

wurDevTim:
If the authentication failed `irods.session.iRODSSession` returns without raising an error. 
The checks in `_get_irods_session` where removed during the merge with RUG, this PR puts them back. 

The function `session.pool.get_connection()` is not ideal as it leaves an open session. 
Instead I tested: `_ = session.server_version`, this function returns the version of the server if it successfully connected, if not it raises an error.

### DIFF
--- a/utils/IrodsConnector.py
+++ b/utils/IrodsConnector.py
@@ -318,7 +318,7 @@ class IrodsConnector():
             try:
                 # Check for good authentication and cache PAM password
                 if 'password' in options:
-                    self._session.pool.get_connection()
+                    _ = self._session.server_version
                     self._write_pam_password()
             except (irods.exception.CAT_INVALID_AUTHENTICATION, KeyError) as error:
                 raise error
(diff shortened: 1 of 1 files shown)


utils/IrodsConnector.py
```diff
@@ -318,7 +318,7 @@ def session(self):
             try:
                 # Check for good authentication and cache PAM password
                 if 'password' in options:
-                    self._session.pool.get_connection()
+                    _ = self._session.server_version
```
jpmcfarland, line 321: This `session` property is ideal because it handles the connection appropriately.

Why not store this value for each authentication method to be used for utility and/or informational purposes so it only needs to be accessed once for a given server session?  The `permissions` property currently uses it and the information tab requires it as well.  It can even be a property itself.

wurDevTim: It not clear to me if you mean the old or new session property with 'this', but the `self._session.pool.get_connection()` is not ideal. As it leaves open connections to the server. `self._session.server_version` is a cleaner way to check for a successful login.

Right now, I am trying to develop in small steps, therefore the scope of this PR is limited to make the login work as intended again. Saving the value or doing some fancy logging could be a new feature.

jpmcfarland: Indeed!  I meant store the results of the test call so it can be reused:

```python
self.server_version = self._session.server_version
```

instead of:

```python
_ = self._session.server_version
```

chStaiger (approved):
Looks very fine to me!!! Will merge!

PR #63, opened 2023-03-11
Irods manager

wurDevTim:
This is going to be one messy PR, so far I have only rewritten the irodsconnector to a ‘manager’ design. 
Still to do are:
- Transform the anonymous into an anonymous manager. 
- Updating all references to the irodsconnector.

It’s my first attempt at such a big rewrite so I am curious to hear what you think. Does it have potential, how would you tackle this challenge? 

Flake8 and pylint trigger on nearly every line, therefore I updated the formatting and rewrote small parts.


chStaiger:
I have not gone through everything but here is a thought. I would an iBridges session which contains all objects that we need later on for the tools:

**IbridgesSession**
*Input*
- ~/.irods/irods_environment*.json (selectable)
- ~/.ibridges/ibridges_config.json (fixed)
- password: either cached in .IrodsA or as user input

*Properties*
- Davrods server: bridges_config
- force_unknown_free_space: ibridges_config
- iRODS session: irods_envirnment.json
- Tabs: GUI tabs
- irods_env_file
- ienv: combined dictionary from irods_emvironment.json and ibridges_config.json
- default_resc: irods_env file
- Password
- Session: iRODS session
- resources (CS: not sure here, can also be moved to the resources class)
- icommands:  True if commands are on OS

*Methods*
- _get_irods_session
- everything to deal with passwords

The instantiated object contains a valid iRODS session and all parameters for the GUI and CLI. Optionally we could also leave those parameters only in the ienv (combined dictionary from irods_emvironment.json and ibridges_config.json)

This object we can then pass to all other classes e.g. the resources, permissions, data_objects etc.

chStaiger:
I think this would also solve this issue: https://github.com/chStaiger/iBridges-Gui/issues/68

PR #69, opened 2023-03-13
[73] Fix Linux crash

jpmcfarland:
Fix the cause of the Linux crash and make minor adjustments along the way.


jpmcfarland:
WRONG BASE!

Issue #65, opened 2023-03-13
utils/utils.py Python 3.6.9 compatibility


maartenschermer:
utils/utils.py is currently uncompatible with Python 3.6.9.
Initial error:
```
  File "/home/maarten/Documents/Projects/ibridges/iBridges-Gui/utils/utils.py", line 4
    from __future__ import annotations
    ^
SyntaxError: future feature annotations is not defined
```

jpmcfarland:
That can be removed.  It was an attempt at more complete type hinting which is not working.

PR #71, opened 2023-03-15
[#65] Remove unneeded annotations

jpmcfarland:
These annotations were an attempt to enhance type hinting that failed.  Since they are incompatible with some versions of supported Python, they are removed.  They can be revisited at a later time if desired.


chStaiger:
Many thanks!

PR #72, opened 2023-03-19
Irods connector rewrite part 1

wurDevTim:
Heey All,

As I feared this rewrite kept growing. It started with splitting the class and adding an interface. But then I noticed all the calls to ‘ic.session.xx’ from the GUI classes, the bugs in the CLI, the partial removal of ‘icommands’ from iBridges.py and a partial factory implementation.  

At the same time this branch is already too big. Therefore I decided not to fix any of the broken stuff outside of the ‘irodsconnector’, The calls to the irods connector are updated so it should behave the same as it did before the rewrite. But be advised that I might have missed a call somewhere in the GUI files. 

@jpmcfarland to implement the lazy instantiation the password and ienv file have to be stored in the (for lack of better name) manager class. To implement this neatly the password getter, setter, delete functions would also have to be placed here. I am not sure if that’s desired. On top of this the other classes have a reference to the session object, and the ‘session’ property does not check if the current session is valid. This can be added but would make the PR even bigger. Therefore I decided to leave it out for now. 

All the other tips and tricks have been added, along with a few new properties and simple functions to replace the calls to 'ic.session.xx'


maartenschermer:
Hi,
re the renaming of some of the functions in manager.py, I feel it's a little confusing that functions that are decorated as properties have "function-like" names. The names of 'get_host', 'get_port', 'get_server_version', 'get_username' suggest that they can be called as function, while they can only be accessed as property. I think 'host', 'port', 'server_version' and  'username' are clearer names.

jpmcfarland:
> @jpmcfarland to implement the lazy instantiation the password and ienv file have to be stored in the (for lack of better name) manager class. To implement this neatly the password getter, setter, delete functions would also have to be placed here. I am not sure if that’s desired. On top of this the other classes have a reference to the session object, and the ‘session’ property does not check if the current session is valid. This can be added but would make the PR even bigger. Therefore I decided to leave it out for now. 

Understood.  I will have a look and request adjustments only where critical, putting the bulk into issues that can focus the effort outside of this PR.

wurDevTim:
Good point, the property calls are renamed. 
And that's a great solution John, I will also add a tickets for rewriting the anonymous client and broken functionalities I found during the rewrite.

iBridgesCli.py
```diff
@@ -168,31 +164,31 @@ def setupIRODS(config, operation):
...
             print(config['iRODS']['irodsresc'] + " upload capacity, free space: No  inofrmation")
     except (NoResultFound, ResourceDoesNotExist):
         print(RED+'iRODS resource does not exist: '+config['iRODS']['irodsresc']+DEFAULT)
         try:
-            print(YEL+'Checking env-file: '+ic.default_resc)
-            resource = ic.get_resource(ic.default_resc)
-            config['iRODS']['irodsresc'] = ic.default_resc
+            print(YEL+'Checking env-file: '+ic.default_resc())
+            resource = ic.get_resource(ic.default_resc())
+            config['iRODS']['irodsresc'] = ic.default_resc()
```
jpmcfarland, line 176: `default_resc` is now a property returning an uncallable string.

wurDevTim: Yep, I saw. This was already the case in the current development branch: 
https://github.com/chStaiger/iBridges-Gui/blob/a2ce3d5033d0caff5521fc2c00d4b121a243c4e2/utils/IrodsConnector.py#L133-L142
Maarten is working on an update for the CLI to fix these issues. 
To save us from doing double I did not fix it here.

jpmcfarland: Okay.

jpmcfarland (approved):

chStaiger (approved):
I tested all functionality from the GUI and CLI.

PR #78, opened 2023-03-28
Ibridges context

chStaiger:
I created an IbridgesContext class which by default uses the `.ibridges/config.json` for all parameters that are not standard variables in the irods_environment.json files.
The irods_environment.json is not known from the beginning and hence instantiated as the empty dictionary and its path as the empty string.
Once the irods_environment.json is known it checks for mandatory keys and also checks whether the irods environment contains keys that actually belong to the ibridges con. This part can be deleted later when we cleared out everything.

To connect the context object to the rest of the GUI I combined both dictionaries into one and pass that to the mailmenu.py:
```py
162             stacked_envs = {}
163             stacked_envs.update(self.context.irods_env)
164             stacked_envs.update(self.context.ibridges_env)
165             browser = gui.mainmenu(widget, conn, stacked_envs)
```

I also commented out the "save config" function in mainmenu.py since that would store the full environment (irods + ibridges). This will be replaced with a new functionality to edit the two config files (https://github.com/chStaiger/iBridges-Gui/issues/49)


jpmcfarland (commented):
Everything looks fine and appears to run fine, but I don't understand the idea behind the two copies of `IbridgesContext` with different contents.  Can I assume that the version in IbridgesContext.py is the prototype and that in `utils.py` will be the final version?

wurDevTim:
Hopefully its the other ways around, `utils.py` already contains to many classes. 
The only thing I don't understand is why you have a `self.ibridges_config` if all the parameters are stored in `self.ibridges_config.config`. Is it an idea to replace `self.ibridges_config.config` with `self.config`?

One way to ensure the same config is used everywhere is to pass a reference to each class at the startup. 
Alternatively you could use a singleton, its not an ideal solution. But does ensures there can only be one instance of the class within the program: https://refactoring.guru/design-patterns/singleton

jpmcfarland:
> Hopefully its the other ways around, `utils.py` already contains to many classes.

That sounds good to me, though I would choose for simpler names such as `context.py` and `class Context`.  We already know it's for iBridges ;p

chStaiger:
I did not intend to also push the utils/IbridgesContext.py
I added the class to utils because it is dependent on the JsonConfig class and if I import that I create circular imports in the utils folder. So at the moment I do not see a straightforward way to create a separate file for the context class.

jpmcfarland:
Also, though redundant, the name `~/.ibridges/ibridges_config.json` mirrors the iRODS naming (`~/.irods/irods_environment.json`) nicely.

jpmcfarland (approved):

chStaiger:
> > Hopefully its the other ways around, `utils.py` already contains to many classes.
> 
> That sounds good to me, though I would choose for simpler names such as `context.py` and `class Context`. We already know it's for iBridges ;p

I adjusted that. Also in the Context class we are now always looking for a .ibridges/ibridges_config.json.

wurDevTim (approved):

chStaiger:
> Hopefully its the other ways around, `utils.py` already contains to many classes. The only thing I don't understand is why you have a `self.ibridges_config` if all the parameters are stored in `self.ibridges_config.config`. Is it an idea to replace `self.ibridges_config.config` with `self.config`?

We in fact have two configs in the whole object, irods and abridges, and we need to make a distinction between them. I will have a lok how to make that more clear.

wurDevTim:
There are a few tricks to create a singleton, what you use now is one of them. 
We are limited to python 3.6 now, so we cannot use the cache decorator (available from python 3.9): https://docs.python.org/3/library/functools.html

There are many possible singleton implementations, the decorator is a nice one.
It uses no additional variables or checks which could cause problem.

PR #79, opened 2023-03-28
Test hooking

maartenschermer:
Moved the Elab-functions out of the main CLI class, introduced mechanism for hooking in plugins, and hooked in the Elab-functions as plugin.


wurDevTim:
@maartenschermer great work moving all the elab connector to to a new class. There was also quite some duplication between the CLI and GUI, do you plan on rewriting the GUI call as well? If not, is it than an idea to make a ticket for this so the work becomes visible?

wurDevTim:
I like that the plugin wraps around the elabConnector,
And the `gui/elabUpload.py` is already loaded on the start, it should be easier to make that optional. 

One other thing I noticed is that pylint and flake8 mark quite some lines. 
The type hinting is also missing, I don't know if you want to add it now, but it would definitely help the next one who works on the class.

maartenschermer:
You're right. I've now corrected all mypy, pylint and flake8 issues where possible.

Issue #73, opened 2023-03-20
Adjust session handling to work better with the reworked connector and GUI classes


jpmcfarland:
- move `ic.session.foo()` calls to appropriate methods in the connector classes (or only manager class?)
- reevaluate where session functionality (data, properties, etc.) is best placed (session.py, manager.py, elsewhere?)
- ensure the current session validity is checked consistently

jpmcfarland:
Writing of the password cache seems broken somehow, but nowhere obvious.

jpmcfarland:
> Writing of the password cache seems broken somehow, but nowhere obvious.

While `_ = self._session.server_version` checks the connection, it also releases it in a way that removes the stored PAM password(s) in `pam_pw_negotiated`, giving nothing to cache locally.  I'm not sure of the exact mechanism, but adding or replacing with `pool.get_connection()` fixes it.

maartenschermer:
It seems that when `session::_write_pam_password()` is called (from here: https://github.com/chStaiger/iBridges-Gui/blob/develop/irodsConnector/session.py#L251), `self._session.pam_pw_negotiated` return an empty list, and the password-cache file isn't written as a result (https://github.com/chStaiger/iBridges-Gui/blob/develop/irodsConnector/session.py#L335).
I'm not familiar with the inner workings of the iRods-library, but I assume that `self._session.pam_pw_negotiated` is called too early, at a point where the underlying variable hasn't been set yet. When I explicitly call `_write_pam_password` (again) on the destruction of the IrodsConnector-class (in `manager.py`), `self._session.pam_pw_negotiated` _does_ return a list, and the cache file _does_ get written.

jpmcfarland:
> An additional issue: even if the password cache file is present, there is currently no way of telling the IrodsConnector() to use it, because the condition `... and password != ''` in this line: https://github.com/chStaiger/iBridges-Gui/blob/develop/irodsConnector/manager.py#L40 prohibits the program to reach the point where the cache file is read (https://github.com/chStaiger/iBridges-Gui/blob/develop/irodsConnector/session.py#L177).

I will have a look.

chStaiger:
@jpmcfarland  Can this issue be closed?

jpmcfarland:
Yes!

PR #80, opened 2023-03-30
[73] Adjust session handling

jpmcfarland:
See Issue #73.  Start with new Context class to store a single copy of configurations and iRODS session.

### DIFF
--- a/gui/IrodsBrowser.py
+++ b/gui/IrodsBrowser.py
@@ -255,7 +255,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget, gui.ui_files.tabBrowser.Ui_tabBrowse
             self.previewBrowser.append(preview_string)
         elif self.ic.dataobject_exists(obj_path):
             obj = self.ic.get_dataobject(obj_path)
-            file_type = utils.utils.IrodsPath(obj_path).suffix[1:]
+            file_type = utils.path.IrodsPath(obj_path).suffix[1:]
             if file_type in ['txt', 'json', 'csv']:
                 try:
                     with obj.open('r') as objfd:
@@ -306,7 +306,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget, gui.ui_files.tabBrowser.Ui_tabBrowse
         try:
             self._clear_error_label()
             self._clear_view_tabs()
-            obj_path = utils.utils.IrodsPath(self.inputPath.text())
+            obj_path = utils.path.IrodsPath(self.inputPath.text())
             if self.ic.collection_exists(obj_path):
                 coll = self.ic.get_collection(obj_path)
                 self.collTable.setRowCount(len(coll.data_objects)+len(coll.subcollections))
@@ -360,7 +360,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget, gui.ui_files.tabBrowser.Ui_tabBrowse
                     "iRODS NETWORK ERROR: No Connection, please check network")
 
     def set_parent_path(self):
-        current_path = utils.utils.IrodsPath(self.inputPath.text())
+        current_path = utils.path.IrodsPath(self.inputPath.text())
         self.inputPath.setText(current_path.parent)
         self.loadTable()
 
@@ -373,7 +373,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget, gui.ui_files.tabBrowser.Ui_tabBrowse
         self._clear_error_label()
         row = index.row()
         obj_path, obj_name = self._get_object_path_name(row)
-        full_path = utils.utils.IrodsPath(obj_path, obj_name)
+        full_path = utils.path.IrodsPath(obj_path, obj_name)
         if self.ic.collection_exists(full_path):
             self.inputPath.setText(full_path)
             self.loadTable()
@@ -388,7 +388,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget, gui.ui_files.tabBrowser.Ui_tabBrowse
         row = index.row()
         self.current_browser_row = row
         obj_path, obj_name = self._get_object_path_name(row)
-        obj_path = utils.utils.IrodsPath(obj_path, obj_name)
+        obj_path = utils.path.IrodsPath(obj_path, obj_name)
         self._clear_view_tabs()
         try:
             self._fill_preview_tab(obj_path)
@@ -544,7 +544,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget, gui.ui_files.tabBrowser.Ui_tabBrowse
         self.errorLabel.clear()
         errors = {}
         obj_path, obj_name = self._get_object_path_name(self.current_browser_row)
-        obj_path = utils.utils.IrodsPath(obj_path, obj_name)
+        obj_path = utils.path.IrodsPath(obj_path, obj_name)
         user_name = self.aclUserField.text()
         if not user_name:
             errors['User name'] = None
--- a/iBridges.py
+++ b/iBridges.py
@@ -5,7 +5,6 @@
 import logging
 import os
 import setproctitle
-import subprocess
 import sys
 
 import irods.exception
@@ -18,24 +17,29 @@ import gui
 from irodsConnector.manager import IrodsConnector
 import utils
 
+APPLICATION_NAME = 'iBridges'
+context = utils.context.Context()
+
 app = PyQt6.QtWidgets.QApplication(sys.argv)
 widget = PyQt6.QtWidgets.QStackedWidget()
 
 # Work around a PRC XML issue handling special characters
 os.environ['PYTHON_IRODSCLIENT_DEFAULT_XML'] = 'QUASI_XML'
 
+
 class IrodsLoginWindow(PyQt6.QtWidgets.QDialog, gui.ui_files.irodsLogin.Ui_irodsLogin):
     """Definition and initialization of the iRODS login window.
 
     """
-
-    this_application = 'iBridges'
+    icommands = False
+    this_application = APPLICATION_NAME
 
     def __init__(self):
         super().__init__()
-        self.icommands = False
+        self.ibridges_path = utils.path.LocalPath(utils.context.IBRIDGES_DIR).expanduser()
+        self.irods_path = utils.path.LocalPath(utils.context.IRODS_DIR).expanduser()
         self._load_gui()
-        self._init_configs_and_logging()
+        self._init_logging()
         self._init_envbox()
         self._init_password()
 
@@ -53,23 +57,11 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog, gui.ui_files.irodsLogin.Ui_irods
         self.ticketButton.clicked.connect(self.ticket_login)
         self.passwordField.setEchoMode(PyQt6.QtWidgets.QLineEdit.EchoMode.Password)
 
-    def _init_configs_and_logging(self):
+    def _init_logging(self):
         """
 
         """
-        # iBridges configuration
-        ibridges_path = utils.utils.LocalPath(
-            os.path.join('~', '.ibridges')).expanduser()
-        if not ibridges_path.is_dir():
-            ibridges_path.mkdir(parents=True)
-        # Loads ibridges config if present, otherwise instantiaties ibridges context with {}
-        self.context = utils.utils.Context()
-        self.irods_path = utils.utils.LocalPath(
-            os.path.join('~', '.irods')).expanduser()
-        if not self.irods_path.is_dir():
-            self.irods_path.mkdir(parents=True)
-        # iBridges logging
-        utils.utils.setup_logger(ibridges_path, 'iBridges')
+        utils.utils.setup_logger(self.ibridges_path, APPLICATION_NAME)
 
     def _init_envbox(self):
         """Populate environment drop-down.
@@ -83,8 +75,8 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog, gui.ui_files.irodsLogin.Ui_irods
         self.envbox.clear()
         self.envbox.addItems(env_jsons)
         envname = ''
-        if 'last_ienv' in self.context.ibridges_env and self.context.ibridges_env['last_ienv'] in env_jsons:
-            envname = self.context.ibridges_env['last_ienv']
+        if 'last_ienv' in context.ibridges and context.ibridges['last_ienv'] in env_jsons:
+            envname = context.ibridges['last_ienv']
         elif 'irods_environment.json' in env_jsons:
             envname = 'irods_environment.json'
         index = 0
@@ -137,14 +129,14 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog, gui.ui_files.irodsLogin.Ui_irods
 
         """
         irods_env_file = self.irods_path.joinpath(self.envbox.currentText())
-        # TODO expand JsonConfig usage to all relevant modules
-        self.context.read_irods_config(irods_env_file)
-        if self.context.irods_env is None:
+        context.irods_env_file = irods_env_file
+        # This should never happen.
+        if context.irods is None:
             self.passError.clear()
             self.envError.setText('ERROR: iRODS environment file not found.')
             self.setCursor(PyQt6.QtGui.QCursor(PyQt6.QtCore.Qt.CursorShape.ArrowCursor))
             return
-        if not utils.utils.can_connect(self.context.irods_env['irods_host']):
+        if not utils.utils.can_connect(context.irods['irods_host']):
             logging.info('iRODS login: No network connection to server')
             self.envError.setText('No network connection to server')
             self.setCursor(PyQt6.QtGui.QCursor(PyQt6.QtCore.Qt.CursorShape.ArrowCursor))
@@ -154,14 +146,14 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog, gui.ui_files.irodsLogin.Ui_irods
         try:
             conn = IrodsConnector(irods_env_file=irods_env_file, password=password)
             # Add own filepath for easy saving.
-            self.context.update_ibridges_keyval('ui_ienvFilePath', irods_env_file)
-            self.context.update_ibridges_keyval('last_ienv', irods_env_file.name)
+            context.ibridges['ui_ienvFilePath'] = irods_env_file
+            context.ibridges['last_ienv'] = irods_env_file.name
             # Save iBridges config to disk and combine with iRODS config.
-            self.context.save_ibridges_config()
-            # widget is a global variable
+            context.save_ibridges()
             stacked_envs = {}
-            stacked_envs.update(self.context.irods_env)
-            stacked_envs.update(self.context.ibridges_env)
+            stacked_envs.update(context.irods)
+            stacked_envs.update(context.ibridges)
+            # widget is a global variable
             browser = gui.mainmenu(widget, conn, stacked_envs)
             if len(widget) == 1:
                 widget.addWidget(browser)
@@ -215,7 +207,7 @@ def main():
     """Main function
 
     """
-    setproctitle.setproctitle('iBridges')
+    setproctitle.setproctitle(APPLICATION_NAME)
     login_window = IrodsLoginWindow()
     widget.addWidget(login_window)
     widget.show()
--- a/irodsConnector/AnonymousManager.py
+++ b/irodsConnector/AnonymousManager.py
@@ -11,9 +11,12 @@ import irods.collection
 from irods.exception import CAT_SQL_ERR
 from irods.session import iRODSSession
 from irods.ticket import Ticket
-from utils.utils import ensure_dir, save_irods_env
 from irodsConnector.Icommands import IrodsConnectorIcommands
 import irodsConnector.keywords as kw
+
+import utils
+
+context = utils.context.Context()
 
 
 # TODO
@@ -52,14 +55,15 @@ class IrodsConnectorAnonymous:
         self.path = path
 
         if IrodsConnectorIcommands.icommands():
-            ensure_dir(os.path.expanduser('~'+os.sep+'.irods'))
+            utils.utils.ensure_dir(os.path.expanduser('~'+os.sep+'.irods'))
             # move previous iRODS sessions to tmp file (envFile and .irodsA file)
             self._move_prev_session_configs(False)
             env = {"irods_host": self.session.host,
                    "irods_port": 1247,
                    "irods_user_name": "anonymous",
                    "irods_zone_name": self.session.zone}
-            save_irods_env(env)
+            context.irods.update(env)
+            context.save_irods()
             logging.info('Anonymous Login: '+self.session.host+', '+self.session.zone)
             pros = Popen(['iinit'], stdout=PIPE, stdin=PIPE, stderr=PIPE, shell=True)
             _, err_login = pros.communicate()
--- a/irodsConnector/session.py
+++ b/irodsConnector/session.py
@@ -1,14 +1,16 @@
 """ session operations
 """
-from json import load
-from os import environ
 import logging
+import json
+import os
 import ssl
+
 import irods.exception
 import irods.password_obfuscation
 import irods.session
 import irodsConnector.keywords as kw
-from utils import utils
+
+import utils
 
 
 class Session(object):
@@ -65,10 +67,10 @@ class Session(object):
 
         """
         if not self._ienv:
-            irods_env_file = utils.LocalPath(self._irods_env_file)
+            irods_env_file = utils.path.LocalPath(self._irods_env_file)
             if irods_env_file.is_file():
                 with open(irods_env_file, encoding='utf-8') as envfd:
-                    self._ienv = load(envfd)
+                    self._ienv = json.load(envfd)
         return self._ienv
 
     @property
@@ -168,10 +170,10 @@ class Session(object):
 
         """
         if not self._password:
-            irods_auth_file = environ.get(
+            irods_auth_file = os.environ.get(
                 'IRODS_AUTHENTICATION_FILE', None)
             if irods_auth_file is None:
-                irods_auth_file = utils.LocalPath(
+                irods_auth_file = utils.path.LocalPath(
                     '~/.irods/.irodsA').expanduser()
             if irods_auth_file.exists():
                 with open(irods_auth_file, encoding='utf-8') as authfd:
--- a/pyinstaller_script.py
+++ b/pyinstaller_script.py
@@ -23,7 +23,7 @@ def run_cmd(command: str):
 
     """
     # POSIX
-    if utils.utils.is_posix():
+    if utils.path.is_posix():
         proc = subprocess.run(
             command, stderr=subprocess.STDOUT, shell=True,
             universal_newlines=True, executable='/bin/bash')
@@ -53,9 +53,9 @@ def ui_to_py(dirname: str, py_exec: str):
         python -m PyQt6.uic.pyuic -x file.ui -o file.py
 
     """
-    for uipath in sorted(utils.utils.LocalPath(dirname).glob('*.ui')):
+    for uipath in sorted(utils.path.LocalPath(dirname).glob('*.ui')):
         basename = uipath.stem
-        pypath = utils.utils.LocalPath(dirname, f'{basename}.py')
+        pypath = utils.path.LocalPath(dirname, f'{basename}.py')
         print(f'Converting {uipath} to {pypath}')
         run_cmd(f'{py_exec} -m PyQt6.uic.pyuic -x {uipath} -o {pypath}')
         # Find and replace the pixelmap references.  They are hardcoded.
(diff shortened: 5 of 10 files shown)


jpmcfarland:
This last commit contained many little changes all over the place.  Hopefully it is not too difficult to follow.

The placeholder for the session in context is an iRODS session not an iBridges session.  I ran into some kind of recursive issue with that and using that here would ignore all the convenience methods in the iBridges session anyway.

Ultimately, if we include session functionality in the singleton context, it would be far better to use the manager itself so that all the functionality is available.  This would mean that the current IrodsConnector class need not be passed as an argument everywhere.  This could unfortunately lead to potential collisions between threads.  Locking is always a possibility.

Comments are quite welcome on this last point.

jpmcfarland:
Trying to reconnect after `Close session` is completely broken.  I'll fix it very soon.

jpmcfarland:
> Trying to reconnect after `Close session` is completely broken. I'll fix it very soon.

Should be fixed.

jpmcfarland:
I'm not sure if all the design decisions are the most ideal, but everything works and the singleton context with connection manager is _VERY_ handy.  I wonder how well this will stand up to multiple threads.  I have some ideas on how to harden it for that if necessary.

jpmcfarland:
Hmm, reconnecting after `Close session` doesn't go through the full process.  It appears to be (re)using the same connection and not a new one.

wurDevTim:
We are already using the same irods connector from multiple threads at the same time during an up/download. So far this works without problem.

jpmcfarland:
> We are already using the same irods connector from multiple threads at the same time during an up/download. So far this works without problem.

That's good news!  If the iRODSSession can handle the multiple threads, the context/connector can.  The singleton instance is created in iBridges.py before any other threads have started, so that should be fine.

jpmcfarland:
> > Hmm, reconnecting after `Close session` doesn't go through the full process. It appears to be (re)using the same connection and not a new one.
> 
> Looking at this, the current implementation of the singleton context is incompatible with a session reset. Minimally, both the iRODS environment and connection manager will need to be effectively reset without creating new instances. I'm checking to see if this is feasible.

I have a strong feeling that the only solution will be to take the Context singleton _out_ of the global namespace so it refreshes more naturally when opening a new session.  There are multiple options.  I'm trying to find the one that works well.

jpmcfarland:
I believe it is finally ready.  @chStaiger, @wurDevTim, please have a look.

jpmcfarland:
These last changes do not affect the main functionality of the PR.  They are minor adjustments primarily related to the unit tests.

chStaiger:
I found some strange things.
My irods_environment_test.json contains the line `"irods_default_resource": "irodsResc"`
But I receive that on the prompt:
```
staig001@busoni iBridges-Gui % ./iBridges.py
Missing key(s) in iRODS environment: ['irods_default_resource']
Please fix and try again!
```

And the irods_connector does not detect any storage resources any longer. It seems that the context class only uses irods_environment.json and no other environment file. I could also not find out how one can pass another irods environment path to the class. @jpmcfarland Would you have a little code snippet for me to test that?

jpmcfarland:
> I found some strange things. My irods_environment_test.json contains the line `"irods_default_resource": "irodsResc"` But I receive that on the prompt:
> 
> ```
> staig001@busoni iBridges-Gui % ./iBridges.py
> Missing key(s) in iRODS environment: ['irods_default_resource']
> Please fix and try again!
> ```
> 
> And the irods_connector does not detect any storage resources any longer. It seems that the context class only uses irods_environment.json and no other environment file. I could also not find out how one can pass another irods environment path to the class. @jpmcfarland Would you have a little code snippet for me to test that?

That _is_ quite strange!

I have the same keyword in environment files on both my test machines.  Can you slack me your file so I can try it out?  Perhaps there is something special in it.

The default name is indeed the default iRODS name, but it should choose what the GUI gives it.  Perhaps this was caused by the first strangeness?  I will have a look.

chStaiger:
I am not sure whether Context should always first try to read in the default irods_environment.json. The file might not be present or might contain insufficient information for ibridges and the Context class. So you will always get logging messages like:
```
Missing key(s) in iRODS environment: ['irods_host', 'irods_user_name', 'irods_port', 'irods_zone_name', 'irods_default_resource']
```
Even though the user has not done anything.

jpmcfarland:
> I am not sure whether Context should always first try to read in the default irods_environment.json. The file might not be present or might contain insufficient information for ibridges and the Context class. So you will always get logging messages like:
> 
> ```
> Missing key(s) in iRODS environment: ['irods_host', 'irods_user_name', 'irods_port', 'irods_zone_name', 'irods_default_resource']
> ```
> 
> Even though the user has not done anything.

Very good point.  I'll see what I can do to adjust that.

jpmcfarland:
It looks like it will indeed use the one chosen connecting when actually connecting.  The default will always be `ibridges_configuration['last_ienv']` if it exists, the iRODS default filename if it doesn't.  This is chosen in `iBridges.py`.

Unfortunately, there is a place that still tries to load a non-existent file, or not finding a file, sets `ibridges_configuration = {}`.  I will see what I can do there.

chStaiger:
More of a general question. Do we want attributes to print information on the prompt or just return the property or None if not set?

jpmcfarland:
> It looks like it will indeed use the one chosen connecting when actually connecting. The default will always be `ibridges_configuration['last_ienv']` if it exists, the iRODS default filename if it doesn't. This is chosen in `iBridges.py`.
> 
> Unfortunately, there is a place that still tries to load a non-existent file, or not finding a file, sets `ibridges_configuration = {}`. I will see what I can do there.

I got rid of the concept of a default iRODS environment file name.  The user is responsible for creating it, so we don't need to test for that specific one, only find JSON files with the correct form in the default iRODS directory.  That fixes the main problem, but now something else is has come up, so I need to fix that before committing.

chStaiger:
I tried to have a closer look at the relationships between ContextContainer, Context, irodsConnector and session.
I find that very confusing. 
The context gets an object `irods_connector`. This one contains an object session which contains again an object `context`.
How does the session get access to the information captured in the context of the ContextContainer?

jpmcfarland:
The `Context` class discovers and provides the iBridges configuration and the iRODS environment.  It also provides a place to store an `IrodsConnector` instance, but leaves it to be set externally (from within iBridges.py).  The job of the `ContextContainer` class is just that: contain the useful parts of a Context instance.  It provides that functionality to any class needing it via inheritance.  Thus, any class inheriting from it gains 4 attributes (`conf`, `conn`, `context`, `ienv`).  `context` is _the_ singleton and the other three are simply shortcuts to the configuration and environment dictionaries and the connector instance.

In iBridges.py, _the_ singleton `Context` instance is created in `main()` and initialized with an `IrodsConnector` instance.  The _magic_ is that no new `Context` instance is created.  Its `__new__()` methods always reuses the same one once set the first time.  This is the case for _all_ class instances whose class inherits from it.  This includes a (recursive?) reference to the `IrodsConnector` instance within the `Resource` and `Session` class instances.  I don't believe any problems will arise from this as accessing one of their methods/properties via the `conn` property is simply a pointer to that very instance.  In any case, it is mainly the iRODS environment used in those two.

I hope this makes it clear.

jpmcfarland:
> @jpmcfarland Yes thank you! So whenever we use the backend (CLI, scripts, new UIs etc) we need to watch out that we create context object before the irodsConnector. I think we need to document that somewhere ;)

Once the first instance of `Context` is made, the next step is setting the `conn` property.  See the bottom of iBridges.py.  Any new startup script (e.g., iBridgesCli.py) will need to create those instances.  Also, any new _plugins_, once we have that ability, will need to inherit from the `ContextContainer` to have access to its payload.

Indeed, we will need to document our API for the plugins when we have them.  This will naturally be a part of that documentation.  We can even start that already with the relevant parts of this conversation put into a README file.

jpmcfarland:
I just discovered that the iRODS environment completeness check is ignored after a session reset.  I'll need to fix that.

jpmcfarland:
> I just discovered that the iRODS environment completeness check is ignored after a session reset. I'll need to fix that.

Fixed!

chStaiger:
I do have some worrying observations when looking at the current backend. I followed the code in the iBridges.py

When I init the context and like we do there:
```
In [1]: from irodsConnector.manager import IrodsConnector
In [2]: from utils.context import Context
In [3]: context = Context()
In [4]: context.irods_connector = IrodsConnector()
```
I can get an infinite procession of `session.conn`, which I am not sure is very handy
In [5]: context.irods_connector.session.conn.session.conn
Out[5]: <irodsConnector.manager.IrodsConnector at 0x107067850>
```

Another thing that worries me is that we can have empty properties at different layers of the objects that we create that make testing, whether some settings are parsed correctly very difficult. E.g. getting the host from a session in an iRODS connector can lead to an unexpected exception, which we would have to check for every time we use the irods connector

```
In [6]: context.irods_connector.session.host
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 context.irods_connector.session.host

File ~/git-repos/iBridges-Gui/irodsConnector/session.py:75, in Session.host(self)
     65 @property
     66 def host(self) -> str:
     67     """Retrieve hostname of the iRODS server
     68
     69     Returns
   (...)
     73
     74     """
---> 75     return self._session.host

AttributeError: 'NoneType' object has no attribute 'host'
```

I think we need to brainstorm about which classes should do some checks on the session object and also on the context class members.

**Comment by jpmcfarland** [comment]
> I do have some worrying observations when looking at the current backend. I followed the code in the iBridges.py
> 
> When I init the context and like we do there:
> 
> ```
> In [1]: from irodsConnector.manager import IrodsConnector
> In [2]: from utils.context import Context
> In [3]: context = Context()
> In [4]: context.irods_connector = IrodsConnector()
> ```
> 
> I can get an infinite procession of `session.conn`, which I am not sure is very handy

This is easily fixed by manually adding the context and select convenience properties to IrodsConnector, Session, and Resource such that they do not inherit from ContextContainer.  This is only done now to avoid code duplication with the unfortunate side-effect you have pointed out.

**Comment by jpmcfarland** [comment]
> Another thing that worries me is that we can have empty properties at different layers of the objects that we create that make testing, whether some settings are parsed correctly very difficult. E.g. getting the host from a session in an iRODS connector can lead to an unexpected exception, which we would have to check for every time we use the irods connector
> 
> ```
> In [6]: context.irods_connector.session.host
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> Input In [6], in <cell line: 1>()
> ----> 1 context.irods_connector.session.host
> 
> File ~/git-repos/iBridges-Gui/irodsConnector/session.py:75, in Session.host(self)
>      65 @property
>      66 def host(self) -> str:
>      67     """Retrieve hostname of the iRODS server
>      68
>      69     Returns
>    (...)
>      73
>      74     """
> ---> 75     return self._session.host
> 
> AttributeError: 'NoneType' object has no attribute 'host'
> ```

This is caused simply by me being sloppy.  A check for the population of the shadow attribute should be included in all these properties.  I only have added these checks where I had errors.  I can go through and update this throughout.

Do you think that will solve it or do you think there is a more fundamental issue?

**Comment by jpmcfarland** [comment]
I have updated the `irodsConnector/*py` properties.  They were generally fine, but needed a little attention.

**Comment by jpmcfarland** [comment]
@chStaiger have your current concerns been addressed by the recent commits?

**Comment by chStaiger** [comment]
> @chStaiger have your current concerns been addressed by the recent commits?

I still have the problem that the GUI does not detect the resources correctly:
```
    -=WARNING=-        -=WARNING=-        -=WARNING=-        -=WARNING=-
The default resource (irodsResc) not found in available resources!
Check "irods_default_resource" and "force_unknown_free_space" settings.
    -=WARNING=-        -=WARNING=-        -=WARNING=-        -=WARNING=-
```
The status tab also does not show any resources.

Although, when I add in line 153 in iBridges.py
```
print(self.conn.resources)
```
I get the full list of resources.

**Comment by chStaiger** [comment]
> > @chStaiger have your current concerns been addressed by the recent commits?
> 
> I still have the problem that the GUI does not detect the resources correctly:
> 
> ```
>     -=WARNING=-        -=WARNING=-        -=WARNING=-        -=WARNING=-
> The default resource (irodsResc) not found in available resources!
> Check "irods_default_resource" and "force_unknown_free_space" settings.
>     -=WARNING=-        -=WARNING=-        -=WARNING=-        -=WARNING=-
> ```
> 
> The status tab also does not show any resources.
> 
> Although, when I add in line 153 in iBridges.py
> 
> ```
> print(self.conn.resources)
> ```
> 
> I get the full list of resources.

Can it be that, the resource.py is ignoring all resources when there is no `free_space` set? 
```
105                 if metadata['parent'] is None and metadata['free_space'] > 0:
106                     resc_names.append(name)
```

**Comment by jpmcfarland** [comment]
@wurDevTim Please check if your iRODS session creation has been improved after the last commit.  I have removed extraneous keywords so only those from the iRODS environment file should be passed to the iRODSSession constructor.

**Comment by jpmcfarland** [comment]
> Can it be that, the resource.py is ignoring all resources when there is no `free_space` set?
> 
> ```
> 105                 if metadata['parent'] is None and metadata['free_space'] > 0:
> 106                     resc_names.append(name)
> ```

Yes, but this check is informational only.  I can have the warning be seen only when the `force_unknown_free_space` flag is False to avoid confusion.  Also, this code is actually at lines 119 and 120.  Please check yours.

**Comment by jpmcfarland** [comment]
Sorry for the disruption, but the only way I found to _properly_ reconcile the logic for resource checking was to split the `force_unknown_free_space` option into `check_free_space` and `force transfers` options.  I'm sure some additional tweaking will be necessary in the functionality.

Please feel free to comment.

**Comment by wurDevTim** [comment]
> @wurDevTim Please check if your iRODS session creation has been improved after the last commit. I have removed extraneous keywords so only those from the iRODS environment file should be passed to the iRODSSession constructor.

(https://github.com/chStaiger/iBridges-Gui/blob/731348b3e9958e72fd3ea93f67d947edf89082c6/irodsConnector/session.py#L286)
misses the _ = options.pop('irods_env_file')

After this change we can connect to the servers at SURF SARA again. 
But the irodsbrowsers remains empty, when I look in the data operaions call from 
https://github.com/chStaiger/iBridges-Gui/blob/731348b3e9958e72fd3ea93f67d947edf89082c6/gui/IrodsBrowser.py#L307
I see that you are now passing an utils.path.IrodsPath, the path exists on the server and is correct.   

image

**Comment by jpmcfarland** [comment]
> > @wurDevTim Please check if your iRODS session creation has been improved after the last commit. I have removed extraneous keywords so only those from the iRODS environment file should be passed to the iRODSSession constructor.
> 
> (
> 
> https://github.com/chStaiger/iBridges-Gui/blob/731348b3e9958e72fd3ea93f67d947edf89082c6/irodsConnector/session.py#L286
> 
> )
> misses the _ = options.pop('irods_env_file')
> After this change we can connect to the servers at SURF SARA again.

Will moving the `pop()` outside of the conditional do the job?  For instance:

```python
        irods_env_file = options.pop('irods_env_file')
        if 'password' not in options:
            try:
                print('AUTH FILE SESSION')
                session = irods.session.iRODSSession(
                    irods_env_file=irods_env_file)
```

wurDevTim:
@jpmcfarland, that works. It returns a valid session.
Do you also have an empty irods browser?

jpmcfarland:
> But the irodsbrowsers remains empty, when I look in the data operaions call from
> 
> https://github.com/chStaiger/iBridges-Gui/blob/731348b3e9958e72fd3ea93f67d947edf89082c6/gui/IrodsBrowser.py#L307
> 
> 
> I see that you are now passing an utils.path.IrodsPath, the path exists on the server and is correct.
> image

All `*Path` objects are `str` objects and `IrodsPath`s are additionally `iRODSPath` objects (which are also `str` objects).  They _MUST_ be compatible!

I will do some testing to see if I can figure it out.  As a work-around, just cast back to `str` for now.

wurDevTim:
> > But the irodsbrowsers remains empty, when I look in the data operaions call from
> > https://github.com/chStaiger/iBridges-Gui/blob/731348b3e9958e72fd3ea93f67d947edf89082c6/gui/IrodsBrowser.py#L307
> > 
> > I see that you are now passing an utils.path.IrodsPath, the path exists on the server and is correct.
> > image
> 
> All `*Path` objects are `str` objects and `IrodsPath`s are additionally `iRODSPath` objects (which are also `str` objects). They _MUST_ be compatible!
> 
> I will do some testing to see if I can figure it out. As a work-around, just cast back to `str` for now.

Too help figure out where to look: the issue is new in this branch. The irods browser is populated correctly on the development branch.

wurDevTim:
@jpmcfarland, after some debugging we discovered that the code works on Ubuntu, not on windows. The cause is most likely the different wrapping of the pathlib library. We noticed that you removed the `is_posix`, and the different implementations for windows and Ubuntu/IOS.

jpmcfarland:
> @jpmcfarland, that works. It returns a valid session.
> Do you also have an empty irods browser?

I need to check.  I just got back from a minor auto pech.

jpmcfarland:
> @jpmcfarland, after some debugging we discovered that the code works on Ubuntu, not on windows. The cause is most likely the different wrapping of the pathlib library. We noticed that you removed the `is_posix`, and the different implementations for windows and Ubuntu/IOS.

While going back through the path module, I realized that the POSIX check was already done by the `pathlib` library itself and I was simply repeating it.  `PurePath`s and `LocalPath`s should be compatible with the platform by design.

wurDevTim:
> ```python
> str(utils.path.IrodsPath("/npec/home/mobile")) == utils.path.IrodsPath("/npec/home/mobile")
> ```
returns False, 
image

I have some plants in the greenhouse which keep growing so I first have to fix some things there. When that is done I will take a look at pathlib. I am also curious about what pathlib is transmitting to the irods server, I have an idea of how to figure that out.

jpmcfarland:
> > ```python
> > str(utils.path.IrodsPath("/npec/home/mobile")) == utils.path.IrodsPath("/npec/home/mobile")
> > ```
> 
> returns False, image
> 

I see enough in your screen capture to convince me it is a library difference issue.  I'll see if I can find a fix or at least a proper workaround for Windows.

jpmcfarland:
> > ```python
> > str(utils.path.IrodsPath("/npec/home/mobile")) == utils.path.IrodsPath("/npec/home/mobile")
> > ```
> 
> returns False, image

It is _effectively_ a PRC bug under Windows, but only when using an `IrodsPath` object, thus _our_ bug:

```python
print(str(utils.path.IrodsPath('/a/b/c')))
/a/b/c
print(repr(str(utils.path.IrodsPath('/a/b/c'))))
'/a/b/c'
print(irods.path.iRODSPath(utils.path.IrodsPath('/a/b/c')))
/\a\b\c
print(repr(irods.path.iRODSPath(utils.path.IrodsPath('/a/b/c'))))
'/\\a\\b\\c'
```

In the `get()` method of a `CollectionManager` the input path is "normalized".  It is the same process the second two commands above illustrate.  Instead of the query receiving `[Collection.name == '/a/b/c']`, it gets `[Collection.name == '/\\a\\b\\c']`, causing the missing collections.

The workaround is to cast `IrodsPath`s to `str`s, but the whole point of the `IrodsPath` object is to prevent this need in the first place!  It effectively turns a `pathlib.Path` object into a `str` object with extra functionality.

Let me see if I can update `IrodsPath` to fix this.  Not likely, but worth a try.

jpmcfarland:
Here is the problem: `p = '/'.join(paths).split('/')` (line 63 of irods/path/__init__.py).  When encountering an `IrodsPath` of ours, its string part is converted from a `pathlib.PurePosixPath` to a `pathlib.PureWindowsPath` during the `join()` and the split sees no `/`s so the expected split does not occur.

```python
paths = (utils.path.IrodsPath("/a/b/c"),)
p = '/'.join(paths).split('/')
print(p)
['', '\\a\\b\\c']
```

Removing the strict POSIX treatment of the `*Paths` is the ultimate cause.  Restoring this treatment fixes the problem:

```python
paths = (utils.path.IrodsPath("/a/b/c"),)
p = '/'.join(paths).split('/')
print(p)
['', 'a', 'b', 'c']
```

d-w-moore:
@jpmcfarland - is `util.path.IrodsPath` a passthru for (or perhaps a copy of)  `irods.path.iRODSPath`?  Forgive my intrusion here, I was just curious ... the present incarnation of iRODSPath in PRC has reverted all POSIX (or indeed OS dependent) usages and relies strictly on string handling to do its work, fwiw : )  I think it was for similar reasons that I had to do this reversion.....

jpmcfarland:
> @jpmcfarland - is `util.path.IrodsPath` an alias for (or perhaps a copy of) `irods.path.iRODSPath`? Forgive my intrusion here, I was just curious ... the present incarnation of `iRODSPath` in the Python iRODS Client has reverted all POSIX (or indeed OS dependent) usages and relies strictly on string handling to do its work, fwiw : ) I think it was for similar reasons that I had to do this reversion.....

Hi @d-w-moore, not a problem!

The similarity is in name only.  I have created a `PurePath` class that is a `str` and uses a `pathlib.PurePath` shadow property to mirror `pathlib` functionality but be seen as a `str` instance.  An `IrodsPath` and `LocalPath` derive from it.  Additionally, `IrodsPath` derives from `irods.path.iRODSPath` as well for _its_ custom path normalization.  `pathlib` has its own path normalization, but it is not quite the same.

There are checks in the PRC that expect a `str`, but a `pathlib.PurePath` does not derive from `str`, so in order to have pathlib functionality without having to cast it as a `str` before passing to a PRC function, an approach like this was necessary.  Unfortunately, it took alot longer to sort out the little details like the one I reverted back to above, but it has been very educational.

d-w-moore:
@jpmcfarland - well, that's certainly some black belt Python right there : )  .  I'd heard of `pathlib` too, but at the time I wrote the `iRODSPath` class, I think I hadn't the depth of understanding to see what advantages it might offer.

d-w-moore:
If it helps, I do know that `__eq__` gets called from the most-derived class, so if  you have `class PathObj(str)`, then `"a"==PathObj("a")`  and `PathObj("a")=="a"` both evaluate to `True`, and both call `PathObj.__eq__` ... but you probably already knew that : )

jpmcfarland:
Indeed!

I certainly stick to the inherited methods when they are doing the correct thing, as I believe they are here.  Unfortunately, the problem above is not with the comparison, but that the path string given to `str.__new__(path.__str__())` is from the wrong type of `pathlib` path.  It is _inherently incorrect_.  Why it is fine on the initial instance but not on the new instance (created by `join()`, I believe), I don't fully understand (Python Voodoo, perhaps?).  I just make sure it is always the correct type now.

chStaiger:
I did a fresh pull this afternoon and found that the writing of the password does not work. Testing shows that
this if statement always evaluates to False:

(session.py)
```
227             try:
228                 # Check for good authentication and cache PAM password
229                 if 'password' in options:
230                     _ = self._session.server_version
231                     self._write_pam_password()
232             except (irods.exception.CAT_INVALID_AUTHENTICATION, KeyError) as er    ror:
233                 raise error
```
 That is caused by `_get_irods_session(options)`

```
284         else:
285             password = options.pop('password')
```

Is it handy to move the password writing outside if the `if`?

README.md
```diff
@@ -45,41 +45,43 @@ python3.10 -m pip install -r requirements.txt
...
-    "ui_tabs": [ 
-        "tabUpDownload", 
-        "tabELNData", 
-        "tabDataBundle", 
-        "tabCreateTicket" 
-    ], 
-    "force_unknown_free_space": false 
+    "check_free_space": true,
+    "davrods_server": "https://server.fqdn.nl",
+    "force_transfers: false",
```
chStaiger, line 64:
```suggestion
    "force_transfers": false",
```

chStaiger (approved):

jpmcfarland:
@chStaiger In this _last_ commit, I removed the extra session validation because it was redundant.  `_get_irods_session()` _always_ returns a validated session or `None`.

jpmcfarland:
@wurDevTim Just wondering how close you are to finalizing your testing.  The longer I look at it the more little non-critical, but ultimately necessary tweaks I find.

wurDevTim:
@jpmcfarland I saw the commits and the nice discussion with Moore, but because it's a big PR I need to find some time to really look at it. In general I am noticing that we don't test enough, which results in broken functionality on the development branch. This week I also found some related to the session, on monday I blocked a few hours to test your branch again on a windows machine. 

Unfortunately that's the earliest moment for me, there is too much going on at my end at the moment.

jpmcfarland:
No worries!  I just wanted to check.  I have other efforts to keep me more than busy enough at the moment.  Thanks for the update!

wurDevTim:
The pyinstaller script was broken in previous changes, I fixed them in order to test your changes in the paths. That looks good now. I still have an issue with the PAM password caching but Christine also tested it on her servers and there it works, so that's probably something in our config. 

In the logging its a little confusing that you always try to start a 'PAM' session first, even if 'PAM' is not in the irods keywords or supported by the server.

It's also not clear to me why you removed the underscores from the private variables? I thought that is how we are supposed to do it in python?

jpmcfarland:
> In the logging its a little confusing that you always try to start a 'PAM' session first, even if 'PAM' is not in the irods keywords or supported by the server.

The PAM check has always been first with the fallback.  I integrated it pretty much as is into the connector.  I can certainly update the logic so that it uses the `irods_authentication_scheme` keyword to dictate the session creation method, but that should definitely be a separate issue/pull request.

> It's also not clear to me why you removed the underscores from the private variables? I thought that is how we are supposed to do it in python?

Part was for consistency with other modules, part was to distinguish between normal local variables and shadow variables used with properties.  For instance:

```python
class A:
    _name = None

    @property
    def name(self):
        if self._name is None:
            # do something
        return self._name
```

If there is a variable that should not be accessed from outside a module that easily could, indicating it private with a leading underscore is the correct method.  I believe they should generally be used sparingly, however.  I will look into some documentation of best practices for using private naming.

jpmcfarland:
> I still have an issue with the PAM password caching but Christine also tested it on her servers and there it works, so that's probably something in our config.

Weird!  I have no problems on all three different OSes.  The ~/.irods/.irodsA file is cached and used properly on them all.  I have just updated some code in that area.  Was this issue seeen before or after that change?

jpmcfarland:
> > In the logging its a little confusing that you always try to start a 'PAM' session first, even if 'PAM' is not in the irods keywords or supported by the server.
> 
> The PAM check has always been first with the fallback. I integrated it pretty much as is into the connector. I can certainly update the logic so that it uses the `irods_authentication_scheme` keyword to dictate the session creation method, but that should definitely be a separate issue/pull request.

Looking into this, I see that if no password is given, the auth file is used for authentication.  If one is given, the contents of the `irods_environment.json` file are sent as keyword arguments to the session constructor with no priority for any authentication scheme.  A `PlainTextPAMPasswordError` is excepted in the case this fails.  No other specific `Exception`s are captured and a set of SSL settings are added to the keywords when the one named `Exception` is captured, and the session creation is attempted for the last time.

I see no code indicating a PAM session attempt.  What errors are you getting and at what stage?

wurDevTim:
> > I still have an issue with the PAM password caching but Christine also tested it on her servers and there it works, so that's probably something in our config.
> 
> Weird! I have no problems on all three different OSes. The ~/.irods/.irodsA file is cached and used properly on them all. I have just updated some code in that area. Was this issue seeen before or after that change?

Its very strange, when you just added the functionality, right after 'the big merge', it worked. We are still using version 1.1.6 of the python irods-client. But we did upgrade our servers from 4.2.11 to 4.3.0. And now the pam_password from a valid session is wrong:
```    
def _write_pam_password(self):
        """Store the returned PAM/LDAP password in the iRODS
        authentication file in obfuscated form.

        """
        connection = self._session.pool.get_connection()
        pam_passwords = self._session.pam_pw_negotiated
```
f24c1c15c1d16e7e937* I changed a few of the numbers to ensure it cannot be decrypted. but left the letters, the decryption of the python irod-client looks at character 6, 


The suprising thing is that the icommands password caching does work on our servers (if we change the "irods_authentication_scheme" from "pam" to  "pam_password". 
Christine told me she is sill using irods v 4.2.11 and for her the python password caching works. which makes me think it's an `irods problem`, not something in this PR. 

P.s. I saw the obfuscation code was updated in python irods-client v1.1.7, but its still responding the same.

jpmcfarland:
> Christine told me she is sill using irods v 4.2.11 and for her the python password caching works. which makes me think it's an `irods problem`, not something in this PR.

Indeed.  We should somehow annotate full compatibility only up to iRODS 4.2.11.

d-w-moore:
> The suprising thing is that the icommands password caching does work on our servers (if we change the "irods_authentication_scheme" from "pam" to "pam_password".
> Christine told me she is sill using irods v 4.2.11 and for her the python password caching works. which makes me think it's an irods problem, not something in this PR.

Not to distract from this thread, but I wanted to say that, if you find investigation is needed into the handling of PAM logins in PRC especially regarding iRODS 4.2.x, let us know - probably best via an email to `support@irods.org`. Note v1.1.8 of the PRC is due to be released soon, with as yet no PAM-related changes planned, but I'm of course interested in knowing specific and compelling evidence of any problems in that area!

d-w-moore:
I understand! thanks @wurDevTim

wurDevTim (approved):
The basics functionality works on windows.

jpmcfarland:
It is bizarre that the conversations only showed up here _after_ the approval.  I have updated them nonetheless.

PR #81, opened 2023-04-05
Ibridges context & config editor

chStaiger:
This is the first version of the config editor on the login screen. It uses the Context class and I would like to wait until @jpmcfarland has finished his work before I adjust my code to the newer version and merge. But you can already test the new feature in this branch.


jpmcfarland:
> This is the first version of the config editor on the login screen. It uses the Context class and I would like to wait until @jpmcfarland has finished his work before I adjust my code to the newer version and merge. But you can already test the new feature in this branch.

That will not be today, unfortunately, but should be by the end of the week.

wurDevTim:
Okay, I will postpone my review untill you finished the work.  In the meantime I will to figure out why the logger broke for me(on the development branch)

jpmcfarland (commented):
It took me a few moments to figure out how to edit values.  in addition to the documentation in the README, perhaps a tiny reminder on the window itself to "double-click to edit a value" might be a good idea.

The text shown in the editor is derived from Python (i.e., text taken from a Python dictionary: True, False, ',  etc.) not from the serialized JSON string (true, false, ", etc.).  This _could_ be confusing to both user and code and result in parse errors or unexpected behavior.

Also, my "irods_connection_pool_refresh_time_in_seconds" was not seen as a valid iRODS keyword.  Looking for a canonical list of keywords on Github, I see: https://github.com/irods/irods/blob/15b28ef46f7c87f701ee11eece17ba838a3c3f00/lib/core/src/irods_configuration_keywords.cpp#L93-L128.  Maybe https://docs.irods.org has something as well.

Perhaps it is a good idea to have individual save buttons for the two configurations.

Also a way to hide the editor could be considered.  It is quite dominant in the window.  Perhaps a separate pop-up triggered by an "Edit configuration" button instead of using the login window itself.

chStaiger:
Converted back to draft. The current version with the new backend is in branch configedit_integration. This PR will stay open until all comments have been taken up in the feature branch.

PR #84, opened 2023-04-26
First try to clean up magic.

chStaiger:
I tried to refactor the irodsConnector/session.py a bit to make it clearer what the dependencies are and also to remove some of the magic that happened by setting attributes. 
Here is my little guideline along which I tried to improve the code:

- only use #property if it is really a property and not in fact a function.
- Passing arguments to functions rather than through inheritance from classes that feel unnatural to be a parent class
- Even if we have a singleton which is handy to make sure that we have one truth, it does not hurt to pass parts of that singleton to functions to make clear what they need and I suspect to also ease the writing of unit tests
- ensure in the constructor that we have a valid and functional object

I also renamed the Session.session to Session.irods_session, which I know will lead to problems when integrating it into the code base, but it makes it much clearer what happens

I would be curious what you think of this style of coding and if you see benefits?

### DIFF
--- a/irodsConnector/session.py
+++ b/irodsConnector/session.py
@@ -15,18 +15,20 @@ import utils
 
 class Session(object):
     """Irods session operations """
-    _session = None
-    context = utils.context.Context()
 
-    def __init__(self, password=''):
+    def __init__(self, context: utils.context.Context, password=None, greet=False):
         """ iRODS authentication with Python client.
 
         Parameters
         ----------
         password : str
             Plain text password.
+        context : utils.context Context object containing valid
+            irods environment and ibridges configuration 
+        greet: bool
+            Print some greeting and irods info to prompt
 
-        The 'password' property can autoload from its cache, but can be
+        The 'password' is autoloaded from its cache, but can be
         overridden by `password` argument.  The iRODS authentication
         file is expected in the standard location (~/.irods/.irodsA) or
         to be specified in the local environment with the
@@ -34,9 +36,43 @@ class Session(object):
 
         """
         self._password = password
+        self._ienv = context.irods_environment.config
+        self._conf = context.ibridges_configuration.config
+        self._irods_env_file = context.irods_env_file
+        self._irods_session = None
+
+        if self._ienv == None:
+            raise ValueError("ERROR SESSION: no irods environment")
+        if self._conf == None:
+            raise ValueError("ERROR SESSION: no ibridges configuration")
+
+        self._irods_session = self._connect_to_irods()
+        self._cache_password()
+        if greet:
+            self.greet()
+
+    def greet(self):
+        """Print irods server info to prompt.
+        """
+
+        print('Welcome to iRODS:')
+        print(f'iRODS Zone: {self._irods_session.zone}')
+        print(f'You are: {self._irods_session.username}')
+        print(f'Default resource: {self.default_resc}')
+        print('You have access to: \n')
+        home_path = f'/{self._irods_session.zone}/home'
+        if self._irods_session.collections.exists(home_path):
+            colls = self._irods_session.collections.get(home_path).subcollections
+            print('\n'.join([coll.path for coll in colls]))
+
+    def _cache_password(self):
+        given_pass = self._password
+        cached_pass = self.read_cached_password
+        if given_pass != cached_pass:
+            self._write_password()
 
     def __del__(self):
-        del self.session
+        self.del_irods_session()
 
     @property
     def conf(self) -> dict:
@@ -48,9 +84,7 @@ class Session(object):
             Configuration from JSON serialized string.
 
         """
-        if self.context.ibridges_configuration:
-            return self.context.ibridges_configuration.config
-        return {}
+        return self._conf
 
     @property
     def davrods(self) -> str:
@@ -62,7 +96,7 @@ class Session(object):
             URL of the configured DavRODS server.
 
         """
-        return self.conf.get('davrods_server', '')
+        return self._conf.get('davrods_server', '')
 
     @property
     def default_resc(self) -> str:
@@ -74,7 +108,7 @@ class Session(object):
             Resource name.
 
         """
-        return self.ienv.get('irods_default_resource', '')
+        return self._ienv.get('irods_default_resource', '')
 
     @property
     def host(self) -> str:
@@ -86,9 +120,9 @@ class Session(object):
             Hostname.
 
         """
-        if self.session:
-            return self.session.host
-        return ''
+        if self._irods_session:
+            return self._irods_session.host
+        return None
 
     @property
     def ienv(self) -> dict:
@@ -99,9 +133,7 @@ class Session(object):
         dict
             Environment from JSON serialized string.
         """
-        if self.context.irods_environment:
-            return self.context.irods_environment.config
-        return {}
+        return self._ienv
 
     @property
     def port(self) -> str:
@@ -113,9 +145,9 @@ class Session(object):
             Port.
 
         """
-        if self.session:
-            return str(self.session.port)
-        return ''
+        if self._irods_session:
+            return str(self._irods_session.port)
+        return None
 
     @property
     def username(self) -> str:
@@ -127,9 +159,9 @@ class Session(object):
             Username.
 
         """
-        if self.session:
-            return self.session.username
-        return ''
+        if self._irods_session:
+            return self._irods_session.username
+        return None
 
     @property
     def server_version(self) -> tuple:
@@ -141,9 +173,9 @@ class Session(object):
             Server version: (major, minor, patch).
 
         """
-        if self.session:
-            return self.session.server_version
-        return ()
+        if self._irods_session:
+            return self._irods_session.server_version
+        return None
 
     @property
     def zone(self) -> str:
@@ -155,12 +187,11 @@ class Session(object):
             Zone.
 
         """
-        if self.session:
-            return self.session.zone
-        return ''
+        if self._irods_session:
+            return self._irods_session.zone
+        return None
 
-    @property
-    def password(self) -> str:
+    def read_cached_password(self) -> str:
         """iRODS password.
 
         Returns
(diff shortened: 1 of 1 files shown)


irodsConnector/session.py
```diff
@@ -15,28 +15,64 @@
...
         file is expected in the standard location (~/.irods/.irodsA) or
         to be specified in the local environment with the
         IRODS_AUTHENTICATION_FILE variable.
 
         """
         self._password = password
+        self._ienv = context.irods_environment.config
+        self._conf = context.ibridges_configuration.config
+        self._irods_env_file = context.irods_env_file
+        self._irods_session = None
```
qubixes, line 42:
```suggestion
        self.irods_session = None
```
As far as I understand the irods_session should be a public attribute, so no underscore then.

irodsConnector/session.py
```diff
@@ -74,7 +108,7 @@ def default_resc(self) -> str:
             Resource name.
 
         """
-        return self.ienv.get('irods_default_resource', '')
+        return self._ienv.get('irods_default_resource', '')
 
     @property
     def host(self) -> str:
```
qubixes, line 114: There are a lot of similar property code that checks the session and passes it through if the session exists (not None). I would suggest using the special `__getattr__` method. Something like:

```python
def __getattr__(self, attr: str):
    if attr in ["host", "username", "port", ...]:
        if self.irods_session:
            return getattr(self.irods_session, attr)
        # Might want to throw an error here.
   return object.__getattribute__(self, attr)
```
I haven't tested it, but something along those lines would save a lot of work and make it easier to add new attributes to be passed through.

irodsConnector/session.py
```diff
@@ -170,90 +201,30 @@ def password(self) -> str:
...
+                self._password = irods.password_obfuscation.decode(authfd.read())
+                return self._password
 
     @property
-    def session(self) -> irods.session.iRODSSession:
-        """iRODS session creation.
+    def irods_session(self):
+        if self._irods_session:
+            return self._irods_session
+        return None
```
qubixes, line 217: This just returns the irods_session, so probably unnecessary after renaming the private attribute to the public attribute.

irodsConnector/session.py
```diff
@@ -268,27 +239,31 @@ def _get_irods_session(options):
...
+                irods_session = irods.session.iRODSSession(password=self._password, **self._ienv)
+                _ = irods_session.server_version
+                logging.info(
+                        'IRODS LOGIN SUCCESS: %s, %s, %s', irods_session.username,
+                        irods_session.zone, irods_session.host)
+                return irods_session
             except irods.connection.PlainTextPAMPasswordError as ptppe:
                 print(f'{kw.RED}SOMETHING WRONG WITH THE ENVIRONMENT JSON? {ptppe!r}{kw.DEFAULT}')
-                try:
+                try: #standard SSL context
```
qubixes, line 266:
```suggestion
                try:  #standard SSL context 
```

irodsConnector/session.py
```diff
@@ -303,27 +278,29 @@ def _get_irods_session(options):
                         'ssl_context': ssl_context,
                     }
                     options.update(ssl_settings)
-                    print('RETRY WITH DEFAULT SSL SETTINGS')
-                    session = irods.session.iRODSSession(password=password, **options)
-                    _ = session.server_version
-                    return session
+                    logging.info('----RETRY WITH DEFAULT SSL SETTINGS')
+                    irods_session = irods.session.iRODSSession(password=password, **options)
```
qubixes, line 282: The options and password local variables are gone, so this needs an update.

irodsConnector/session.py
```diff
@@ -303,27 +278,29 @@ def _get_irods_session(options):
...
+                    logging.info('iRODS AUTH FILE SESSION LOGIN FAILED: {error!r}')
                     raise error
             except Exception as autherror:
                 logging.info('AUTHENTICATION ERROR')
-                print(f'{kw.RED}AUTHENTICATION ERROR: {autherror!r}{kw.DEFAULT}')
                 raise autherror
 
-    def _write_pam_password(self):
-        """Store the returned PAM/LDAP password in the iRODS
+    def _write_password(self):
```
qubixes, line 295: Probably nice to have this close to the `cache_password` method.

qubixes (commented):
I think it's a nice start! I have made some additional suggestions to make the class a bit leaner/easier to understand (and some small PEP 8 suggestions).

irodsConnector/session.py
jpmcfarland: Interesting approach, but the developer (and more particularly a user) will need to know a given attribute is available as its value is only returned when asked for and is never _really_ an attribute of this object.  A property is always visible and has the added benefit of custom functionality without resorting to a long list of conditionals in `__getattr__()`.  For instance:

```python
        if attr == 'attr1':
            # do something
        if attr == 'attr2':
            # do domething else
        # etc.
```

It does keep the functionality consolidated, however.  Perhaps I'll come across a use for this at some point.

irodsConnector/session.py
```diff
@@ -170,90 +201,30 @@ def password(self) -> str:
...
-                    colls = self._session.collections.get(home_path).subcollections
-                    print('\n'.join([coll.path for coll in colls]))
-                logging.info(
-                    'IRODS LOGIN SUCCESS: %s, %s, %s', self._session.username,
-                    self._session.zone, self._session.host)
-        return self._session
-
-    @session.deleter
-    def session(self):
+    def del_irods_session(self):
```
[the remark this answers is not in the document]
jpmcfarland: Quite interesting, but we want to keep the session open and allow the `iRODSSession` _context management_ to close connections as needed while maintaining the session until `we` choose to close it.  With this method, we would constantly be creating new sessions.  The idea behind the singleton `Context` object is to house among other things, a single set of configurations and a _single_ iRODS session.

irodsConnector/session.py
```diff
@@ -15,28 +15,64 @@
...
+        The 'password' is autoloaded from its cache, but can be
         overridden by `password` argument.  The iRODS authentication
         file is expected in the standard location (~/.irods/.irodsA) or
         to be specified in the local environment with the
         IRODS_AUTHENTICATION_FILE variable.
 
         """
         self._password = password
+        self._ienv = context.irods_environment.config
+        self._conf = context.ibridges_configuration.config
```
jpmcfarland, line 40: Firstly, if either of these `JsonConfig`s fail to load for any reason, there will be an error something like:

```python
AttributeError: NoneType object has no attribute 'config'
```

Secondly, there is no lazy instantiation benefit here.  The instant you access the `irods_environment` or `ibridges_configuration` property of the **context**, and the `config` property for each of those, you are first instantiating `JsonConfig` objects, then loading their dictionaries from disk.  When this is done as an iBridges `Session` instance is initialized, these four instances are created, and only at that point.  Resetting them after a 'Close session' becomes more cumbersome as a result, requiring extra manual methods.  The `@property`s do this naturally as part of their design.

irodsConnector/session.py
```diff
@@ -15,28 +15,64 @@
...
+        self._irods_env_file = context.irods_env_file
```
jpmcfarland, line 41: Why mirror this variable?  Why not just use it directly?  An independent copy used in only one location could lead to inconsistencies when switching environment files between sessions because the local class variable will not updated when the passed **context** variable is, or vice versa.

irodsConnector/session.py
```diff
@@ -15,28 +15,64 @@
...
+        self._irods_session = None
+
+        if self._ienv == None:
+            raise ValueError("ERROR SESSION: no irods environment")
+        if self._conf == None:
+            raise ValueError("ERROR SESSION: no ibridges configuration")
```
jpmcfarland, line 47: The check for valid configurations is done only once upon initialization here.  As I mentioned for lines 39-40, for a new session, a new instance of this class will be required or separate methods will be needed to reset and check the configurations each time the session is closed.  `@property`s are better suited for this as they have on-demand access to the current state of the **context**.

irodsConnector/session.py
```diff
@@ -15,28 +15,64 @@
...
+
+        self._irods_session = self._connect_to_irods()
+        self._cache_password()
```
jpmcfarland, line 50: Again, these actions take place only once for an instance and must be manually repeated for every new session.

irodsConnector/session.py
```diff
@@ -155,12 +187,11 @@ def zone(self) -> str:
             Zone.
 
         """
-        if self.session:
-            return self.session.zone
-        return ''
+        if self._irods_session:
+            return self._irods_session.zone
+        return None
```
jpmcfarland, line 192: An `iRODSSession` returns strings for all these attributes (they are actually `@property`s of that class) except `server_version`, which returns a tuple.  Returning `None` when the session not available is an inconsistent return value and will require _any and all_ code using these properties to check first if they exist to avoid an `AttributeError`.

irodsConnector/session.py
```diff
@@ -170,90 +201,30 @@ def password(self) -> str:
...
-            del self._session
-            self._session = None
+        if self._irods_session is not None:
+            self._irods_session.cleanup()
+            del self._irods_session
+            self._irods_session = None
 
-    @staticmethod
-    def _get_irods_session(options):
+    def _connect_to_irods(self):
```
jpmcfarland, line 227: I don't understand the need to _manually_ manage the attached `iRODSSession` with separate logical methods.  Simply accessing, assigning, or deleting a `@property` performs these operations automatically.  If a **session** does not exist, the **getter** initiates one, if one does exist, it bypasses that process and simply returns it.  That is the _purpose_ of a `@property`.

I can see that an inadvertent access of a property could cause trouble.  Using a method to check the private shadow variable can always be provided to check its status.  So far, I have found this unnecessary on the whole, but while writing unit tests recently, I see some possible potential need.

Ultimately, it is simplest to allow the object to do the work _for_ you.  Ask it what its state is if you're curious, but let it do what is needed based on that state.  The PRC uses quite a number of such `@property`s, the greatest number in the `session` module, in fact.  I don't see a reason _not_ to use them similarly here when they are beneficial.

qubixes:
> > I tried to refactor the irodsConnector/session.py a bit to make it clearer what the dependencies are and also to remove some of the magic that happened by setting attributes. Here is my little guideline along which I tried to improve the code:
> > 
> > * only use #property if it is really a property and not in fact a function.
> 
> A `@property` is, by definition, a **function**. It is effectively a _smart attribute_ that can behave differently based on the state of the instance of its class. When certain attributes are set in different ways, it behaves accordingly.

Yes, technically they are similar as methods. However, they convey a different expectation to the user/developer, namely the expectation that they behave similar to attributes. I have found the following excerpt in PEP 8:

> Note 1: Try to keep the functional behavior side-effect free, although side-effects such as caching are generally fine.

> Note 2: Avoid using properties for computationally expensive operations; the attribute notation makes the caller believe that access is (relatively) cheap.

Personally, I think creating a session by accessing a property is too much of a side effect, and it might also violate the second note if the server is slow to respond.

> Discussion about passing singletons.

As I mentioned before, I'm not completely convinced that a singleton is the right choice here (but then again I don't know the rest of the code (much)). However, I agree that if you decide on using singletons, it doesn't make all that much sense to pass them around as parameters. What I don't understand is why you would set it as an attribute/base class though. You can access the context from anywhere using `context=user.context.Context()`, right? It seems hard to imagine it would take much time to initiate a singleton.

> The constructor _initializes_ an instance and does so only once. In a system where many instances are created at once, performing validation at the time of instantiation can be costly. Also, when many of these instances are (necessarily) reused and must be validated over and over, on-demand validation using the dynamic nature of properties is far more preferable.

Why do these instances have to be reused over and over? Performance?

irodsConnector/session.py
qubixes: If the attributes are all different, then yes, a property for each makes much more sense.

With visibility, I assume you mean being able to autocomplete in iPython/IDE? Yes, I guess that is a disadvantage (though technically it should be possible for the IDE to resolve it). The code to me just seemed like it was passing through all the attributes from the `irods_session`.  You could also consider passing through *all* attributes from the `irods_session` object, but by that point you're probably better off just inheriting from there..

jpmcfarland:
> > > I tried to refactor the irodsConnector/session.py a bit to make it clearer what the dependencies are and also to remove some of the magic that happened by setting attributes. Here is my little guideline along which I tried to improve the code:
> > > 
> > > * only use #property if it is really a property and not in fact a function.
> > 
> > 
> > A `@property` is, by definition, a **function**. It is effectively a _smart attribute_ that can behave differently based on the state of the instance of its class. When certain attributes are set in different ways, it behaves accordingly.
> 
> Yes, technically they are similar as methods. However, they convey a different expectation to the user/developer, namely the expectation that they behave similar to attributes. I have found the following excerpt in PEP 8:
> 
> > Note 1: Try to keep the functional behavior side-effect free, although side-effects such as caching are generally fine.
> 
> > Note 2: Avoid using properties for computationally expensive operations; the attribute notation makes the caller believe that access is (relatively) cheap.
> 
> Personally, I think creating a session by accessing a property is too much of a side effect, and it might also violate the second note if the server is slow to respond.

I did mean **method** there, but it amounts to the same thing: some code replacing a simple lookup/value return.

As for the expense of the session, I would completely agree in the standard case, particularly if that session is throw-away and needs to be renewed often.  Here, there absolutely _can be_ a great expense (time, not computational) establishing a session, but it is spent only once for each instance created.  The iRODS session is `cached` until 'Close session' is called, and the next 'Connect' will create a new one--but only then!

As for side effects, regardless of whether a property `getter` creates the session or a manually called method does, the side-effects will be _exactly_ the same.  Certain setup _must_ first be done prior to either.  I prefer to avoid the manual call unless there is a need for it.  I just don't see the need in this case.  For me it just adds unnecessary complexity, not clarity.

> 
> > Discussion about passing singletons.
> 
> As I mentioned before, I'm not completely convinced that a singleton is the right choice here (but then again I don't know the rest of the code (much)). However, I agree that if you decide on using singletons, it doesn't make all that much sense to pass them around as parameters. What I don't understand is why you would set it as an attribute/base class though. You can access the context from anywhere using `context=user.context.Context()`, right? It seems hard to imagine it would take much time to initiate a singleton.

Indeed!

Consider, however, there are currently 18 classes that inherit from the `ContextContainer`.  This class has only two jobs: 1) define three convenience properties to shorten the names accessing the main functionality of the `Context` class, and 2) `context = utils.context.Context()`.

> > The constructor _initializes_ an instance and does so only once. In a system where many instances are created at once, performing validation at the time of instantiation can be costly. Also, when many of these instances are (necessarily) reused and must be validated over and over, on-demand validation using the dynamic nature of properties is far more preferable.
> 
> Why do these instances have to be reused over and over? Performance?

Good question.  That need came about originally when I was using globals in each module to access the context and its configuration objects.  I needed a method to avoid reassignment of _all_those globals, retaining the original instances, but performing a required `reset()` between 'Close session' and 'Connect'.  Switching from globals to inheritance does not negate the inherent usefulness of this functionality, so I kept it.  I didn't want to have to develop it again in case it was later required.

PR #86, opened 2023-05-01
[85] Allow independent backend session

jpmcfarland:
Mainly, improve error handling and transparency for the session manager.

### DIFF
--- a/irodsConnector/session.py
+++ b/irodsConnector/session.py
@@ -137,10 +137,7 @@ class Session(object):
                 self._irods_session.cleanup()
             except NameError:
                 pass
-            try:
-                del self._irods_session
-            except AttributeError:
-                pass
+            del self._irods_session
             self._irods_session = None
 
     # Authentication workflow methods
--- a/irodsConnector/session.py
+++ b/irodsConnector/session.py
@@ -9,13 +9,13 @@ import irods.exception
 import irods.password_obfuscation
 import irods.session
 
-from . import keywords as kw
 import utils
+from . import keywords as kw
 
 
 class Session(object):
     """Irods session operations """
-    _session = None
+    _irods_session = None
     # Singleton instance initially configured in iBridges.py
     context = utils.context.Context()
 
@@ -37,7 +37,7 @@ class Session(object):
         self._password = password
 
     def __del__(self):
-        del self.session
+        del self.irods_session
 
     @property
     def conf(self) -> dict:
@@ -87,8 +87,8 @@ class Session(object):
             Hostname.
 
         """
-        if self.session:
-            return self.session.host
+        if self.has_irods_session():
+            return self.irods_session.host
         return ''
 
     @property
@@ -114,8 +114,8 @@ class Session(object):
             Port.
 
         """
-        if self.session:
-            return str(self.session.port)
+        if self.has_irods_session():
+            return str(self.irods_session.port)
         return ''
 
     @property
@@ -128,8 +128,8 @@ class Session(object):
             Username.
 
         """
-        if self.session:
-            return self.session.username
+        if self.has_irods_session():
+            return self.irods_session.username
         return ''
 
     @property
@@ -142,8 +142,8 @@ class Session(object):
             Server version: (major, minor, patch).
 
         """
-        if self.session:
-            return self.session.server_version
+        if self.has_irods_session():
+            return self.irods_session.server_version
         return ()
 
     @property
@@ -156,8 +156,8 @@ class Session(object):
             Zone.
 
         """
-        if self.session:
-            return self.session.zone
+        if self.has_irods_session():
+            return self.irods_session.zone
         return ''
 
     @property
@@ -203,7 +203,7 @@ class Session(object):
         self._password = ''
 
     @property
-    def session(self) -> irods.session.iRODSSession:
+    def irods_session(self) -> irods.session.iRODSSession:
         """iRODS session creation.
 
         Returns
@@ -212,7 +212,7 @@ class Session(object):
             iRODS connection based on given environment and password.
 
         """
-        if self._session is None:
+        if self._irods_session is None:
             if not self.context.irods_env_file:
                 if 'last_ienv' in self.conf:
                     print(f'{kw.YEL}"irods_env_file" not set.  Using "last_ienv" value.{kw.DEFAULT}')
@@ -233,37 +233,37 @@ class Session(object):
             del self.password
             if given_pass != cached_pass:
                 options['password'] = given_pass
-            self._session = self._get_irods_session(options)
+            self._irods_session = self._get_irods_session(options)
             # If session exists, it is validated.
-            if self._session:
+            if self._irods_session:
                 if given_pass != cached_pass:
                     self._write_pam_password()
                 print('Welcome to iRODS:')
-                print(f'iRODS Zone: {self._session.zone}')
-                print(f'You are: {self._session.username}')
+                print(f'iRODS Zone: {self._irods_session.zone}')
+                print(f'You are: {self._irods_session.username}')
                 print(f'Default resource: {self.default_resc}')
                 print('You have access to: \n')
-                home_path = f'/{self._session.zone}/home'
-                if self._session.collections.exists(home_path):
-                    colls = self._session.collections.get(home_path).subcollections
+                home_path = f'/{self._irods_session.zone}/home'
+                if self._irods_session.collections.exists(home_path):
+                    colls = self._irods_session.collections.get(home_path).subcollections
                     print('\n'.join([coll.path for coll in colls]))
                 logging.info(
-                    'IRODS LOGIN SUCCESS: %s, %s, %s', self._session.username,
-                    self._session.zone, self._session.host)
-        return self._session
+                    'IRODS LOGIN SUCCESS: %s, %s, %s', self._irods_session.username,
+                    self._irods_session.zone, self._irods_session.host)
+        return self._irods_session
 
-    @session.deleter
-    def session(self):
-        """Properly delete irods session.
+    @irods_session.deleter
+    def irods_session(self):
+        """Properly delete iRODS session.
         """
-        if self._session is not None:
-            # In case the session is not really there.
+        if self._irods_session is not None:
+            # In case the iRODS session is not fully there.
             try:
-                self._session.cleanup()
+                self._irods_session.cleanup()
             except NameError:
                 pass
-            del self._session
-            self._session = None
+            del self._irods_session
+            self._irods_session = None
 
     @staticmethod
     def _get_irods_session(options):
@@ -336,10 +336,10 @@ class Session(object):
         authentication file in obfuscated form.
 
         """
-        connection = self._session.pool.get_connection()
-        pam_passwords = self._session.pam_pw_negotiated
+        connection = self._irods_session.pool.get_connection()
+        pam_passwords = self._irods_session.pam_pw_negotiated
         if len(pam_passwords):
-            irods_auth_file = self._session.get_irods_password_file()
+            irods_auth_file = self._irods_session.get_irods_password_file()
             with open(irods_auth_file, 'w', encoding='utf-8') as authfd:
                 authfd.write(
                     irods.password_obfuscation.encode(pam_passwords[0]))
@@ -347,7 +347,7 @@ class Session(object):
             logging.info('WARNING -- unable to cache obfuscated password locally')
         connection.release()
 
-    def has_session(self) -> bool:
+    def has_irods_session(self) -> bool:
         """Check if an iRODS session has been assigned to its shadow
         variable.
 
@@ -357,4 +357,4 @@ class Session(object):
             Has a session been set?
 
         """
-        return self._session is not None
+        return self._irods_session is not None
--- a/irodsConnector/session.py
+++ b/irodsConnector/session.py
@@ -196,22 +196,9 @@ class Session(object):
         if self._irods_session:
             if given_pass != cached_pass:
                 self._write_pam_password()
-            self._give_irods_greeting()
-
-    def _give_irods_greeting(self):
-        logging.info('Welcome to iRODS:')
-        logging.info(f'iRODS Zone: {self._irods_session.zone}')
-        logging.info(f'You are: {self._irods_session.username}')
-        logging.info(f'Default resource: {self.default_resc}')
-        logging.info('You have access to: \n')
-        home_path = f'/{self._irods_session.zone}/home'
-        if self._irods_session.collections.exists(home_path):
-            colls = self._irods_session.collections.get(home_path).subcollections
-            for coll in colls:
-                logging.info('\t%s', coll.path)
-        logging.info(
-            'IRODS LOGIN SUCCESS: %s:%s', self._irods_session.host,
-            self._irods_session.port)
+            logging.info(
+                'IRODS LOGIN SUCCESS: %s:%s', self._irods_session.host,
+                self._irods_session.port)
 
     @staticmethod
     def _get_irods_session(options):
--- a/irodsConnector/session.py
+++ b/irodsConnector/session.py
@@ -347,6 +347,13 @@ class Session(object):
             logging.info('WARNING -- unable to cache obfuscated password locally')
         connection.release()
 
+    def connect(self):
+        """Manually establish an iRODS session.
+
+        """
+        if not self.has_irods_session():
+            _ = self.irods_session.server_version
+
     def has_irods_session(self) -> bool:
         """Check if an iRODS session has been assigned to its shadow
         variable.
--- a/irodsConnector/session.py
+++ b/irodsConnector/session.py
@@ -2,7 +2,6 @@
 """
 import logging
 import os
-import ssl
 
 import irods.connection
 import irods.exception
@@ -16,8 +15,9 @@ from . import keywords as kw
 class Session(object):
     """Irods session operations """
     _irods_session = None
-    # Singleton instance initially configured in iBridges.py
-    context = utils.context.Context()
+    ibridges_configuration = None
+    irods_env_file = ''
+    irods_environment = None
 
     def __init__(self, password=''):
         """ iRODS authentication with Python client.
@@ -51,8 +51,9 @@ class Session(object):
             Configuration from JSON serialized string.
 
         """
-        if self.context.ibridges_configuration:
-            return self.context.ibridges_configuration.config
+        logging.debug(f'getting: {self.ibridges_configuration=}')
+        if self.ibridges_configuration:
+            return self.ibridges_configuration.config
         return {}
 
     @property
@@ -64,8 +65,9 @@ class Session(object):
         dict
             Environment from JSON serialized string.
         """
-        if self.context.irods_environment:
-            return self.context.irods_environment.config
+        logging.debug(f'getting: {self.irods_environment=}')
+        if self.irods_environment:
+            return self.irods_environment.config
         return {}
 
     # Authentication workflow properties
@@ -170,17 +172,19 @@ class Session(object):
         """Establish an iRODS session.
 
         """
-        if not self.context.irods_env_file:
+        logging.debug(f'{self.irods_env_file=}')
+        if not self.irods_env_file:
             if 'last_ienv' in self.conf:
                 logging.warning(f'{kw.YEL}"irods_env_file" not set.  Using "last_ienv" value.{kw.DEFAULT}')
                 irods_path = utils.path.LocalPath(utils.context.IRODS_DIR).expanduser()
-                self.context.irods_env_file = irods_path.joinpath(self.conf['last_ienv'])
+                self.irods_env_file = irods_path.joinpath(self.conf['last_ienv'])
             else:
                 logging.error(f'{kw.RED}No iRODS session: "irods_env_file" not set!{kw.DEFAULT}')
                 return
         options = {
-            'irods_env_file': str(self.context.irods_env_file),
+            'irods_env_file': str(self.irods_env_file),
         }
+        logging.debug(f'{self.ienv=}')
         if self.ienv is not None:
             options.update(self.ienv)
         given_pass = self.password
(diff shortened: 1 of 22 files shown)


chStaiger:
I think we need to go a bit further in the redesign of the flow. Something that is unlogical is, that if I ask e.g. for the property `session.host`, the property triggers the authentication workflow, while it would be expected that when there is a valid and authenticated session the property will return the host and if there is no authenticated session something like `None` or `''`. The latter can now be nicely implemented with `session.has_session`.

This is what I did:
```py
import irodsConnector
import utils
context.irods_connector = irodsConnector.manager.IrodsConnector()
context.irods_env_file = "<my file>"
context.irods_connector.password = "<mypassword>"
```
Now I ask for the host (or any other property that depends on the session):
```
In [12]: context.irods_connector.session.host
FULL ENVIRONMENT SESSION
Welcome to iRODS:
iRODS Zone: <zone>
You are: <user>
Default resource: irodsResc
You have access to:

/zone/home/datamanager-its
...
Out[12]: 'hostname'
```

If we separate the properties from the authentication workflow, that also means that we need another trigger or a separate function for the authentication.

Just imagine if you would execute this workflow asking for a simple thing like the hostname and the client would hang there for 20 seconds because of network issues ...

maartenschermer:
There's a double `del self.session` in _manager.py_, one in the destructor, and again in the subsequent `@session.deleter`, causing an `AttributeError` (as `_session` no longer exists by then).

maartenschermer:
I agree with Chistine, as a developer, there's no way to tell when the manager is going to try and connect. When I read out `session.password`, it does not connect, whereas when I access `session.username`, it _does_. This behaviour can probably be explained, and also be documented, but I feel it runs against intuition. As a developer, I feel way more in control if I actively have to call 'connect'.

jpmcfarland:
> If we separate the properties from the authentication workflow, that also means that we need another trigger or a separate function for the authentication.
> 
> Just imagine if you would execute this workflow asking for a simple thing like the hostname and the client would hang there for 20 seconds because of network issues ...

I have to ask, what iBridges workflow requires no iRODS session?  The only meaningful workflow I can think of is testing, but that is a _very_ special case.  There we would use mocks for unit tests and a test server for functional tests.

I could conceive that another possibility is manually experimenting with the code as above.  Do we expect users to be doing this?  And if they do, should they expect a problem free and intuitive "under the hood" experience?  I certainly wouldn't.  I would be using the software in a manner not intended by the author and would expect no end of problems until I understood  the code well enough.  Not something we should expect the average iBridges user would do.  In our case, I expect only us to be doing this kind of access, and we know what to expect.

I think developing the code to support such a specialized use-case may not be the best idea.  We could easily lose sight of the main use-case and end up developing all sorts of `toeters en bellen` along the way.  I know this because I already fell into that trap a bit and perfected some code I previously wrote to add attribute access to a dictionary.  Handy for manually introspecting JsonConfigs, but totally unnecessary!

All this aside, I could easily see using `has_session()` (as you hinted above) to avoid the authentication trigger in these circumstances.  That would mean many changes similar to modifying:

```python
        if self.session:
            return self.session.host
        return ''
```

into:

```python
        if self.has_session:
            return self.session.host
        return ''
```

Also as you said above, that would require manually triggering the authentication when desired instead of when it is first requested.  Is that necessary to appease two very specific, limited use-cases?  I certainly would like to think about what effect that could have on the main use-case.

jpmcfarland:
> I agree with Christine. As a developer there's no way to tell when the manager is going to try and connect. When I read out `session.password`, it does not connect, whereas when I access `session.username`, it _does_. This behaviour can probably be explained, and also be documented, but I feel it runs against intuition. As a developer, I feel way more in control if I actively have to call 'connect'.

The nature of iRODS session properties and those not part of it will always be difficult to disentangle.  Your example here is one of the major ones.  `username` is taken from the iRODS environment file as part of the session creation and we use the established session as the source of its value.  `password`, on the other hand is quite different.  While there is password information obtainable from the iRODS session, the purpose of this property is not informative like `username` (or `host`, `port`, and `zone` for that matter), but derives its initial value from the iRODS authentication file where an obfuscated version is cached.  It can also be set from the login dialog.  It has a more complex use than most properties.

One way to at least partially clarify this is to only take values from the iRODS session when there is no other source (e.g., `server_version`) and take those that exist in the iRODS environment file directly from there.

Is that an idea?

jpmcfarland:
@chStaiger I have added the `connect()` method and integrated it.  I have also differentiated the iBridges and iRODS session properties everywhere.  That was more complex than I initially thought.

For me, the only thing still on the table is the session properties I mentioned above.  Is it preferable to use the environment values or wait until the session is established?

Is there anything else you feel needs addressing?

maartenschermer:
> > There's a double `del self.session` in _manager.py_, one in the destructor, and again in the subsequent `@session.deleter`, causing an `AttributeError` (as `_session` no longer exists by then).
> 
> The destructor calls `del self.session` which is intended to delete the property. This then calls calls the `@session.deleter` which calls `del self._session`. This in turn calls the `iRODSSession` destructor if populated. if `self._session` is None, the `self._session` attribute is deleted (`None` cannot be). In any case, this is followed by `self._session = None`, reinitializing the attribute.
> 
> I don't see where this error could be coming from. Can you post the error and call that leads to it?

If I run this:
`from irodsConnector.manager import IrodsConnector
irods_conn = IrodsConnector("?")
`
I get:
`Exception ignored in: <function IrodsConnector.__del__ at 0x7f0da9cae1f0>
  File "/iBridges-Gui/irodsConnector/manager.py", line 53, in __del__
  File "/iBridges-Gui/irodsConnector/manager.py", line 107, in session
AttributeError: _session
`
Note that this only happens when the destructor is invoked, so when you run the code above in iPython, the error doesn't occur until you exit the iPython shell.

irodsConnector/session.py
```diff
@@ -9,13 +9,14 @@
...
 import utils
+from . import keywords as kw
 
 
 class Session(object):
     """Irods session operations """
-    _session = None
+    _irods_session = None
+    # Singleton instance initially configured in iBridges.py
     context = utils.context.Context()
```
chStaiger, line 20: With that we can still create circular properties, we somehow need to break the circle:
```
context.irods_connector.session.context.irods_connector.session.context.irods_connector...
```
I see two options here:
Let only the manager use the context object and pass parameters for the session through, or do not add the manager object to the context object.

irodsConnector/session.py
```diff
@@ -320,13 +336,32 @@ def _write_pam_password(self):
...
-            irods_auth_file = self._session.get_irods_password_file()
+            irods_auth_file = self._irods_session.get_irods_password_file()
             with open(irods_auth_file, 'w', encoding='utf-8') as authfd:
                 authfd.write(
                     irods.password_obfuscation.encode(pam_passwords[0]))
         else:
             logging.info('WARNING -- unable to cache obfuscated password locally')
         connection.release()
+
+    def connect(self):
```
chStaiger, line 350: What would speak against putting the whole authentication workflow here? 
In case that there is already a valid connection we could
1) cleanup that connection and reinstantiate a new one
2) return the existing session

The attribute irods_session would then also return either a valid connection or None.

jpmcfarland:
> If I run this:
> 
> `from irodsConnector.manager import IrodsConnector` `irods_conn = IrodsConnector("?")`
> 
> I get:
> 
> `Exception ignored in: <function IrodsConnector.__del__ at 0x7f0da9cae1f0>` `Traceback (most recent call last):` ` File "/iBridges-Gui/irodsConnector/manager.py", line 53, in __del__` ` File "/iBridges-Gui/irodsConnector/manager.py", line 107, in session` `AttributeError: _session`
> 
> Note that this only happens when the destructor is invoked, so when you run the code above in iPython, the error doesn't occur until you exit the iPython shell.

Very strange!  It appears that the property deleter is somehow running more than once?  The Python documentation clearly states:

"If `c` is an instance of [a class that `x` is a property of], `c.x` will invoke the getter, `c.x = value` will invoke the setter, and `del c.x` [will invoke] the deleter."  This is for:

```python
    @x.deleter
    def x(self):
        del self._x
```

Exactly _this_ case!  What is even weirder is I tried to check it first, and got the same error:

```python
        if hasattr(self, '_session'):
            del self._session
```

The class _clearly_ has the attribute but still has an error.  The only thing that appears to work is to capture the `AttributError` in a try...except clause.  I will apply that to all deleters for good measure.

irodsConnector/session.py
jpmcfarland: A person goes to the doctor and says, "Doctor, it hurts when I do this."  The doctor responds, "Then don't do that!"

Does this circular reference cause any issues?  Could it lead to issues?  Garbage collection aside, I can't think of any.

For this case, the manager _cannot_ use the context without potentially having the same issue.  Only `resource` and `session` use the configurations.  So the only possibility I see would be to pass those through the manager.  Not a problem, but it feels to me like a step back to be passing around what can easily be obtained via a singleton.

I would be concerned that future functionality creates a need for more dynamic access to the configurations and we would need to revert to the current situation, but I will see if there are currently any negative effects of making this change.  If I don't have any issues I'll commit it.

jpmcfarland: Unfortunately, the session creation involving the _auth file_ that enables cached password logins requires the name of the _environment file_.  This value is normally passed via the context object.  Also, the recent change to allow using the previously cached environment file as the 'last_ienv' value is also affected by this.  That code needs to update the value in the context object.

I can pass this value too (I was afraid it wouldn't be so simple), but the 'last_ienv' clause will need to be removed if the Session class no longer has access to the context object.

What would you prefer?  I'm suspecting it will be dropping the 'last_ienv' convenience to remove the circular reference.

chStaiger: I think we can drop the 'last_ienv' (in fact I have already done so in PR #89 ). It is a nice convenience to set a previous environment, but we can leave that to the front end, where we actually already do that. For the backend I do not see any need, we can throw an error of not enough information was given to instantiate the session.

jpmcfarland: Unfortunately it will not work without far more setup.  Right now, an interactive session can be started with:

```python
import utils
import irodsConnector
context.irods_connector = irodsConnector.manager.IrodsConnector()
```

optionally specifying the `irods_env_file` and possibly a password.  For the resource and session objects, if context is not used to pass the configurations and `irods_env_file` value (which now _must_ be manually set), they will all have to be set manually:

```python
import utils
import irodsConnector
context.irods_connector = irodsConnector.manager.IrodsConnector()
context.irods_env_file = '~/.irods/irods_environment.json'
context.irods_connector.resource.conf = self.context.ibridges_configuration.config
context.irods_connector.resource.ienv = self.context.irods_environment.config
context.irods_connector.session.conf = self.context.ibridges_configuration.config
context.irods_connector.session.ienv = self.context.irods_environment.config
context.irods_connector.session.irods_env_file = context.irods_env_file
```

This is because the iBridges session and resource objects are already instantiated _before_ the `irods_env_file` value and configuration dictionaries can be passed.  They must be updated.  The iBridges configuration is not as problematic, but must be set the same way.  For the GUI, the `context.irods_env_file` is already set, but the appropriate version of the last 5 lines will need to follow that in the code.

All in all, not so nice!

jpmcfarland:
> > I have to ask, what iBridges workflow requires no iRODS session? The only meaningful workflow I can think of is testing, but that is a _very_ special case. There we would use mocks for unit tests and a test server for functional tests.
> > I could conceive that another possibility is manually experimenting with the code as above. Do we expect users to be doing this? And if they do, should they expect a problem free and intuitive "under the hood" experience? I certainly wouldn't. I would be using the software in a manner not intended by the author and would expect no end of problems until I understood the code well enough. Not something we should expect the average iBridges user would do. In our case, I expect only us to be doing this kind of access, and we know what to expect.
> 
> You are right, there is no single feature in the GUI nor the CLI and also testing which does not need an iRODS session. That is why we need to be so sure that it is instantiated and authenticated correctly. From a certain point in the GUI and CLI workflow, we do not want to worry any longer about whether the authentication will be successful. We need an authenticated irods session. That point in the GUI, is when the user clicks the connect button and there is a similar confirm signal that the user sends in the CLI.
> 
> We cannot be sure from the software side when there is enough and valid information given, that is something that only the user of the software knows. So the 'lazy' execution of the authentication workflow does not really fit here. As soon as someone hits the "connect" button we want an authenticated iRODS session or give the user some information what went wrong. Now this is only implicitly done by trying to create the objects in mainmenu.py.

There is nothing _implicit_ about this process.  Clicking "Connect" runs the `login_function()` method in `iBridges.py`.  The environment file and password, from the user input, are added to the configuration.  A connection is explicitly triggered with a `connect()` method.  Within the workflow triggered by this method, the iRODS session is validated.  There IS a validated, authenticated iRODS session at that point, any errors already captured if there is an issue.  Any objects created in `mainmenu.py` will already have an explicitly created, validated, authenticated iRODS session available.

I simply don't understand what's wrong this.

Might it be that _lazy_ instantiation/execution/evaluation with properties is confused with instantiation that is somehow _implicit_; that the property is somehow _hiding_ this functionality or its execution?  It certainly is not.  Take for example the old style of property creation, there is always up to 3 separate methods given as arguments to the `property` constructor.  Decorators are now used, but the structure is effectively the same.  By convention the methods for creating a property are named:

```python
class A:
    _my_property = None

    def _get_myproperty(self):
        # do something
        return self._my_property

    def _set_myproperty(self, value):
        # do something else
        self._my_property = value

    def _del_myproperty(self):
        # do something like:
        del self._my_property

    myproperty = property(_get_myproperty, _set_myproperty, _del_myproperty)
```

These mehtods are marked private with the leading underscore in their names, but there is nothing preventing them being called directly if desired.  There is not even a need to mark them so, it's just convention.  They can just as easily be called: `get_myproperty`, `set_myproperty`, and `del_myproperty`.

The point is that when the value of a property is requested, the `getter` method is run.  Whether run in this way by the property or manually, the result is identical:

```python
# from above
a = A()
a.myproperty == a._get_myproperty()  # the same as getattr(a, 'myproperty')
True
a.myproperty = value  # the same as a._set_myproperty(value) or setattr(a, 'myproperty', value)
del a.myproperty  # the same as a._del_myproperty() or delattr(a, 'myproperty')
```

The property simply packages the three methods into one entity.  What makes it lazy is what the code they contain does.  The way I have written this example, the behavior is identical to `a.myproperty` as a simple attribute.  The utility comes when you add clauses like: `if self._my_property is not None:`.  That is when its functionality is the most useful.

irodsConnector/session.py
jpmcfarland: > The attribute irods_session would then also return either a valid connection or None.  Give that our frontend logic: do not create a new session when there is an authenticated one is already implemented in the manager.py, I think we can dare to be more general here and create a new session as soon as someone asks for it through the function.

I believe the `irods_session` property getter is doing exactly this already.  If no session is established, `self._irods_session` remains `None` when it is returned.  It can certainly be cleaned up a bit, but this is essentially the current situation.

The workflow currently resides in mainly three methods (four if you count the `@irods_session.deleter`).  That is too much code for a single method.  Further modularizing it (splitting up and simplifying `_get_irods_session()`, for instance) and moving the bulk of the _getter_ code to `connect()` certainly cannot hurt.  I believe the `irods_session` property should still be used to house the actual `iRODSSession`.  I just cannot predict its form in this new configuration.

One proposal I have in addition to this is perhaps to simplfy the `try...except` clause around `self.conn.connect()` in `iBridges.py` and instead capture the errors within the session and return them to the `login_function` already parsed.  Then that functionality can be used outside the GUI.

Shall I take a crack at all this to see what I come up with?

jpmcfarland: >  What would speak against putting the whole authentication workflow here?
> In case that there is already a valid connection we could
>
> 1. cleanup that connection and reinstantiate a new one
> 2. return the existing session

Looking back at this, how do we determine the choice between **1.** and **2.**?

Currently, we create an `iRODSSession` with the authentication tokens for the chosen environment.  That session remains until "Close session" is selected.  At that point, the **context** is _reset_ and a whole _new_ `IrodsConnector` and associated _new_ `iRODSSession` is created with the _new_ authentication tokens for the _new_ chosen environment.  The two parts of **1.** are in two different workflows!

Returning the existing, valid session (**2.**) is the job of the `irods_session` property regardless of what else it may or may not do.

As for a valid _connection_, that is the responsibility of the `iRODSSession`.  It has its own `Connection`s, connection `Pool`s, and linked logic to arbitrate a valid _connection_.  The only definitive method to check for a _connected_ session is to query for the `server_version` property.  There are also **idle** and **active** pools holding the actual `Connection`s.  I would not want to use those facts because I don't know what logic (current and future) governs them.

One thing I _can_ do, though it may be overkill (and have performance consequences), is to let the `irods_session` be set by the `connect()` method then check it's _validity_ (get the `server_version`) before returning the _valid_ session instance.  This would happen _every time_ `context.session.irods_session` is accessed ensuring:

> The attribute irods_session would then also return either a valid connection or None.

Will this satisfy?

chStaiger: Please have a look ar PR #89. This is a pull request to this PR/branch which tackles this problem.

jpmcfarland: I had already finished my updates and the main content difference is the `greetings_from_irods()` method.  That is a good idea, but where is it decided to show it or not?

The main functional difference is the validity test for the `irods_session`.  I believe that is important for your goal.

Also, is the removal of the SSL context safe?  I have never really tested it.

Lastly, where do we prefer to retrieve the session introspection values from?  The environment file, the session, or both?

jpmcfarland: Perhaps it is best to add a verbosity flag to the iBridges configuration.  QUIET for no greeting or other messages, NORMAL for the greeting and informational messages, DEBUG for _everything_.  Errors will always be shown.

chStaiger: > I had already finished my updates and the main content difference is the `greetings_from_irods()` method. That is a good idea, but where is it decided to show it or not?
> 
The connect function has a boolean `greet`. Default is `False` but one can set it to `True`.
> The main functional difference is the validity test for the `irods_session`. I believe that is important for your goal.
> 
If the connect function fails it will create an Exception which is caught in the front end and then translated to a more useful error message for the user.  The backend can also not fix missing or wrong rods parameters. So I would say, we do not need extra checking in the backend.
> Also, is the removal of the SSL context safe? I have never really tested it.
It is, it was really just a convenience for YODA users at a time where the SSL context was not yet communicated to the users.
> 
> Lastly, where do we prefer to retrieve the session introspection values from? The environment file, the session, or both?
That is a tricky one. In the session.py, I would tend to retrieve all parameters from the actual irods_session as it is instantiated and authenticated. This would also give us the chance for debugging to compare with the parameters in context which are provided by the config files, just to test whether all parameters are set correctly in the actual irods_session.

jpmcfarland: > > I had already finished my updates and the main content difference is the `greetings_from_irods()` method. That is a good idea, but where is it decided to show it or not?
> 
> The connect function has a boolean `greet`. Default is `False` but one can set it to `True`.

What I meant was, does the user decide whether or not to see this?  How would the user indicate that?  I believe that a "verbose" setting would be the most generic way to handle it.  Python logging has a  number of log levels.  This is clearly an INFO level message.  I can update the logger to output `logging.info()` both to the log file and to standard out.  This allows only one call instead of separate print statements.  The logging messages to the GUI uses a completely different mechanism, so that will need to be dealt with differently.

> > Lastly, where do we prefer to retrieve the session introspection values from? The environment file, the session, or both?
> 
> That is a tricky one. In the session.py, I would tend to retrieve all parameters from the actual irods_session as it is instantiated and authenticated. This would also give us the chance for debugging to compare with the parameters in context which are provided by the config files, just to test whether all parameters are set correctly in the actual irods_session.

I don't think that is a useful exercise.  The PRC is so well developed that these values will likely _always_ be set correctly.  If they weren't, the session would have other noticeable problems.

I think the only reason to use the session values is to give an indication that there _is_ a session.  Using `server_version` is the best way to determine that.  It is just simpler to take the values from the environment where possible.  That's what I would recommend.

irodsConnector/session.py
jpmcfarland: I finally managed the correct method of _passing_ the configurations to the `irodsConnector` classes _after_ instantiation and _prior_ to the iRODS **session** creation in order to remove those pesky circular references.  It appears a little _under-the-table_, but I was already using this method for updating other aspects of the `Context` during a **Close session**.  It works quite well and I hope clear enough.  Testing will show!

The new logging has some issues that i still need figure out and I have not yet tested _interactive_ instantiation of the `Context` and iRODS session.  @chStaiger I'm sure you will let me know where you find problems.

jpmcfarland:
> > If I run this:
> > `from irodsConnector.manager import IrodsConnector` `irods_conn = IrodsConnector("?")`
> > I get:
> > `Exception ignored in: <function IrodsConnector.__del__ at 0x7f0da9cae1f0>` `Traceback (most recent call last):` ` File "/iBridges-Gui/irodsConnector/manager.py", line 53, in __del__` ` File "/iBridges-Gui/irodsConnector/manager.py", line 107, in session` `AttributeError: _session`
> > Note that this only happens when the destructor is invoked, so when you run the code above in iPython, the error doesn't occur until you exit the iPython shell.
> 
> Very strange! It appears that the property deleter is somehow running more than once? The Python documentation clearly states:
> 
> "If `c` is an instance of [a class that `x` is a property of], `c.x` will invoke the getter, `c.x = value` will invoke the setter, and `del c.x` [will invoke] the deleter." This is for:
> 
> ```python
>     @x.deleter
>     def x(self):
>         del self._x
> ```
> 
> Exactly _this_ case! What is even weirder is I tried to check it first, and got the same error:
> 
> ```python
>         if hasattr(self, '_session'):
>             del self._session
> ```
> 
> The class _clearly_ has the attribute but still has an error. The only thing that appears to work is to capture the `AttributeError` in a try...except clause. I will apply that to all deleters for good measure.

@maartenschermer The circular reference just removed from the `Context.irods_connector` appears to have been the cause of this behavior.  Please have a look and verify this has been resolved.

chStaiger:
I think we can remove the irods greetings from the backend. they are at most useful for the command line interface and then they should be implemented there. I will remove them.

chStaiger:
I tested the GUI and everything works fine. From my side this PR is ready to be merged.

jpmcfarland:
> I tested the GUI and everything works fine. From my side this PR is ready to be merged.

WooHoo!!!

PR #88, opened 2023-05-03
[82] Update style, documentation, and tests

jpmcfarland:
Apply PEP 8 guidelines, update documentation, and most importantly improve unit tests.


jpmcfarland:
So far, this PR has the following updates:

- docstrings all over
- naming in iBridges.py, gui/mainmenu.py, and utils/*
- a subtle password bugfix in iBridges.py
- split of utils tesing modules and tons of new tests
- a bugfix for ibridges_configuration and irods_environment properties
  in Context
- formally make ContextContainer abstract
- updates to utils modules as required by new/updated tests

I think this is far more than intended for this PR and pushing manageability.  I think I will redefine the scope of the parent issue a bit and add tasks.  In any case there should be multiple PRs and not just one huge one.

chStaiger (approved):
Gathering all path-functionality in the respective path objects is very nice, I love the tests!!!

I will create a ticket that when we move to the next release that we need to update the docs for developers (create your own tab).

jpmcfarland:
@jtersteegwur Please let me know if you have any reservations.  I am going to resolve the conflicts and merge later today.

jpmcfarland:
@chStaiger What was broken in this branch that was fixed in develop?

chStaiger:
> @chStaiger What was broken in this branch that was fixed in develop?

The most important thing was not to delete the irodsConnector when resetting the context.

jpmcfarland:
I made sure that was updated.  No worries there.

I am having another issue with the password logic in the iBridges/manager/session files where the cached password cannot be discovered because the `irods_environment` is now required to instantiate an iBridges `Session`, but the `irods_env_file` is unknown at the time of instantiation.  I'm working on a solution.

jpmcfarland:
Still working on the solution.  Sorry for the delays.  I have had precious little time the last two weeks!

chStaiger:
> Still working on the solution. Sorry for the delays. I have had precious little time the last two weeks!

No problem, I am meanwhile getting into the dockerisation of iRODS.

jpmcfarland:
@chStaiger Please check iBridgesCli.py functioning with the current changes.  The fix required some logic changes and simplifications and I was unable to fully test that.  Also, _some_ minor recent logic changes needed to be reverted to solve some testing issues that came up.

Everything should be well-oiled now!

chStaiger:
Little bug in context:

```
241     def reset(self):
242         """Reset existing instances of dynamic class members.
243
244         """
245         logging.debug('Resetting Context.')
246         self.irods_connector.reset()
```

If the context does not have an irodsConnector, this will fail.
--> `if self.irods_connector: ....`

chStaiger:
For some reason the iBridgesCli.py fails when one wants to login with a password and not with a cached password.
The manager.session.ienv is not set. I tried to figure out why, but I do not see why. Here is what happens in the code:


```
self.irods_conn = IrodsConnector()
self.irods_conn.password = secret
self.irods_conn.ibridges_configuration = self.context.ibridges_configuration
self.irods_conn.irods_env_file = self.context.irods_env_file
self.irods_conn.irods_environment = self.context.irods_environment
self.irods_conn.connect()
```

I can reproduce that in ipython:

```
import utils
from irodsConnector.manager import IrodsConnector

context.irods_env_file = "/USER/.irods/irods_environment.json"
irods_conn = IrodsConnector()
irods_conn.password = ("shsfghahbhfja")
irods_conn.ibridges_configuration = context.ibridges_configuration
irods_conn.irods_env_file = context.irods_env_file
irods_conn.irods_environment = context.irods_environment
irods_conn.connect()
```

The `session.ienv` is empty `{}` and the connect fails with `TypeError: iRODSAccount.__init__() missing 4 required positional arguments: 'irods_host', 'irods_port', 'irods_user_name', and 'irods_zone_name'`

I can fix that by adding an `irods_conn.reset()` before doing the `irods_conn.connect()`. But that does not seem to be logical. @jpmcfarland what do you think?

chStaiger:
And I think the context is sometimes too lazy ;)

```
In [2]: context = utils.context.Context()

In [3]: context.irods_env_file = "/USER/.irods/irods_environment_home.json"

In [4]: context.irods_environment
Out[4]: JSONConfig("{}")
```

Do we not want to load the json directly when the path is set?

chStaiger (changes requested):
The current instantiation of the session.py does not ensure that the ienv is properly set.

jpmcfarland:
> Little bug in context:
> 
> ```
> 241     def reset(self):
> 242         """Reset existing instances of dynamic class members.
> 243
> 244         """
> 245         logging.debug('Resetting Context.')
> 246         self.irods_connector.reset()
> ```
> 
> If the context does not have an irodsConnector, this will fail. --> `if self.irods_connector: ....`

This is actually _not_ a bug.  By **design**, the `Context` class is assumed _always_ to have an `IrodsConnector` for normal functioning.

I will certainly add a check for existence to avoid the possibility of this error because setting the `Context.irods_connector` property is not enforced in any way.

I would recommend using the `IrodsConnector` instance from _within_ the `Context` singleton for the CLI if possible.  This will keep it consistent with the GUI and avoid potential bugs.

jpmcfarland:
> And I think the context is sometimes too lazy ;)
> 
> ```
> In [2]: context = utils.context.Context()
> 
> In [3]: context.irods_env_file = "/USER/.irods/irods_environment_home.json"
> 
> In [4]: context.irods_environment
> Out[4]: JSONConfig("{}")
> ```
> 
> Do we not want to load the json directly when the path is set and the Json object is asked for?

It appears that setting `context.irods_env_file` does not set the `context.irods_environment.pathname`, but looking at the code, it may be that after a `reset()` the `context.irods_environment` property is checking for the wrong thing.

I test for this scenario specifically, but without the intervening `reset()`.  Certainly an _esoteric bug_.

jpmcfarland:
> > And I think the context is sometimes too lazy ;)
> > ```
> > In [2]: context = utils.context.Context()
> > 
> > In [3]: context.irods_env_file = "/USER/.irods/irods_environment_home.json"
> > 
> > In [4]: context.irods_environment
> > Out[4]: JSONConfig("{}")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do we not want to load the json directly when the path is set and the Json object is asked for?
> 
> It appears that setting `context.irods_env_file` does not set the `context.irods_environment.filepath`, but looking at the code, it may be that after a `reset()` the `context.irods_environment` property is checking for the wrong thing.
> 
> I test for this scenario specifically, but without the intervening `reset()`. Certainly an _esoteric bug_.

Upon investigation, the `JSONConfig.filepath` file is not loaded into the empty dictionary (hence `JSONConfig("{}")`) until the `config` property is accessed .  This is a cosmetic issue with `JSONConfig`, not a functional bug with `Context`.  The former is doing exactly what it is meant to, however unintuitive.

I have updated the `JSONConfig.__repr__()` method to access the `config` property instead of the `_config` shadow variable.  This effectively loads the JSON file if it exists and is not already loaded.  Again, this fix is only cosmetic.

jpmcfarland:
> I can reproduce that in ipython:
> 
```python
import utils
from irodsConnector.manager import IrodsConnector

context.irods_env_file = "/USER/.irods/irods_environment.json"
irods_conn = context.irods_connector = IrodsConnector()
irods_conn.password = ("shsfghahbhfja")
irods_conn.ibridges_configuration = context.ibridges_configuration
irods_conn.irods_env_file = context.irods_env_file
irods_conn.irods_environment = context.irods_environment
irods_conn.connect()
```
> 
> The `session.ienv` is empty `{}` and the connect fails with `TypeError: iRODSAccount.__init__() missing 4 required positional arguments: 'irods_host', 'irods_port', 'irods_user_name', and 'irods_zone_name'`
> 
> I can fix that by adding an `irods_conn.reset()` before doing the `irods_conn.connect()`. But that does not seem to be logical. @jpmcfarland what do you think?

As I said before, this is caused by _not_ setting the `context.irods_connector` property.  If you make the small modification to the `irods_conn =` line as I have above, everything works fine.

jpmcfarland:
> > The current instantiation of the session.py does not ensure that the ienv is properly set.
> 
> That is not its job, though it should certainly check if it is the case. I will sort it out.
> 
> Incidentally, attempting a `Session()` instantiation with a `context.irods_environment.config` when the `context.irods_environment` did not exist was one of the main bugs I found after the merge from develop.

The `irods_environment.config` dictionary is already checked in `iBridges.py` in the `login_function()` method and I have added various checks in the `IrodsConnector` class and an exception in the `Session.connect()` method in case it gets that far.  Will this be sufficient to _ensure_ it is properly set?

jpmcfarland:
@chStaiger The changes are committed.  Please have a look.

chStaiger:
> As I said before, this is caused by _not_ setting the `context.irods_connector` property. If you make the small modification to the `irods_conn =` line as I have above, everything works fine.
> 
> Please test if this works for you. I can even add it as a work-around in this PR. The better solution would involve using `self.context.irods_connector`in the place of `self.irods_conn`.

Does that mean that the irodsConnector is not independent from the context any longer? I do not think that this is a good idea for a design. I thought we introduced the the context for convenience, but not as a necessity?!

jpmcfarland:
> > As I said before, this is caused by _not_ setting the `context.irods_connector` property. If you make the small modification to the `irods_conn =` line as I have above, everything works fine.
> > Please test if this works for you. I can even add it as a work-around in this PR. The better solution would involve using `self.context.irods_connector`in the place of `self.irods_conn`.
> 
> Does that mean that the irodsConnector is not independent from the context any longer? I do not think that this is a good idea for a design. I thought we introduced the the context for convenience, but not as a necessity?!

The `IrodsConnector` has _never_ been independent of the `Context` in the sense that the iRODS session and the iRODS environment are intimately related and dependent.  We just formalized the connection and chose to put them in a single software construct for convenience.

We could invert the relation (`Context` inside `Connector`) or even completely divide the two (passing arguments ad infinitum), but the relationship between the iRODS environment and the iRODS session will _always_ be the same.  It is just a matter of certain convenience for keeping them together.  When this is the case, many things become easier development-wise because they are a single unit, but there are associated restrictions.  The trade-off is similar in the opposite case: greater flexibility at the cost of coding ease.  Both can be made to work.  Both can be made as transparent.  But a choice must be made, and the current design is the choice we all agreed on when we first seriously discussed the `Context`.

The bottom line right now is that there is a place in the `Context` for the `Connector`, and the code design and implementation is _optimized for_ and _takes advantage of_ that arrangement.  If we choose to stay with this and use it as designed, we are basically done.

chStaiger:
> The `IrodsConnector` has _never_ been independent of the `Context` in the sense that the iRODS session and the iRODS environment are intimately related and dependent. We just formalized the connection and chose to put them in a single software construct for convenience.
> 

In the current version of the code one can instantiate the irodsConnector independently from the context by setting the necessary parameters from JsonConfig objects, e.g. . 

```
# load configurations
ibridgesEnv = JsonConfig("/USER/.ibridges/ibridges_config.json")
irodsEnv = JsonConfig("/USER/.irods/irods_environment.json")

# create irods connection
ic = IrodsConnector()
ic.ibridges_configuration = ibridgesEnv
ic.irods_environment = irodsEnv
ic.irods_env_file = irodsEnv.filepath
ic.password = “my password"
ic.connect()
# do something on the iRODS server
```

One can also add the irodsConnector to the singleton Context, which indeed can be handy for the GUI. For the Cli it is not necessary since we have a very linear and straightforward algorithm which implements the up and downloads. 

However, it was not needed for the irodsConnector to be part of the singleton in order to function. And it was also not needed for the context to contain an irodsConnector. To change that and make that mandatory is a bigger design change which the whole development team needs to be aware of and agree upon. The change might also have an influence on other features e.g. the parallelisation of data transfers.
That discussion is a bit too big for this pull request.

> We could invert the relation (`Context` inside `Connector`) or even completely divide the two (passing arguments ad infinitum), but the relationship between the iRODS environment and the iRODS session will _always_ be the same. It is just a matter of certain convenience for keeping them together. When this is the case, many things become easier development-wise because they are a single unit, but there are associated restrictions. The trade-off is similar in the opposite case: greater flexibility at the cost of coding ease. Both can be made to work. Both can be made as transparent. But a choice must be made, and the current design is the choice we all agreed on when we first seriously discussed the `Context`.
> 
> The bottom line right now is that there is a place in the `Context` for the `Connector`, and the code design and implementation is _optimized for_ and _takes advantage of_ that arrangement. If we choose to stay with this and use it as designed, we are basically done.

Is it an idea to separate the password bug and this design question from this PR and put it into a different PR? It would also be nice to have some instructions how to reproduce the password bug. We can then decide in our next meeting which road we want to take to resolve this.

jpmcfarland:
> > The `IrodsConnector` has _never_ been independent of the `Context` in the sense that the iRODS session and the iRODS environment are intimately related and dependent. We just formalized the connection and chose to put them in a single software construct for convenience.
> 
> In the current version of the code one can instantiate the irodsConnector independently from the context by setting the necessary parameters from JsonConfig objects, e.g. .
> 
> ```
> # load configurations
> ibridgesEnv = JsonConfig("/USER/.ibridges/ibridges_config.json")
> irodsEnv = JsonConfig("/USER/.irods/irods_environment.json")
> 
> # create irods connection
> ic = IrodsConnector()
> ic.ibridges_configuration = ibridgesEnv
> ic.irods_environment = irodsEnv
> ic.irods_env_file = irodsEnv.filepath
> ic.password = “my password"
> ic.connect()
> # do something on the iRODS server
> ```
> 
> One can also add the irodsConnector to the singleton Context, which indeed can be handy for the GUI. For the Cli it is not necessary since we have a very linear and straightforward algorithm which implements the up and downloads.
> 
> However, it was not needed for the irodsConnector to be part of the singleton in order to function. And it was also not needed for the context to contain an irodsConnector. To change that and make that mandatory is a bigger design change which the whole development team needs to be aware of and agree upon. The change might also have an influence on other features e.g. the parallelisation of data transfers. That discussion is a bit too big for this pull request.

Whoa!  We _did_ agree upon this exact design _months_ ago.  _All_ of us, I thought.  This has been in place since early April: https://github.com/chStaiger/iBridges-Gui/commit/2bdbc3827ba5a6d6e18ca27823a3c15df9e1a66c.  It is already in `develop` and `master`.  I really don't understand this.

> 
> > We could invert the relation (`Context` inside `Connector`) or even completely divide the two (passing arguments ad infinitum), but the relationship between the iRODS environment and the iRODS session will _always_ be the same. It is just a matter of certain convenience for keeping them together. When this is the case, many things become easier development-wise because they are a single unit, but there are associated restrictions. The trade-off is similar in the opposite case: greater flexibility at the cost of coding ease. Both can be made to work. Both can be made as transparent. But a choice must be made, and the current design is the choice we all agreed on when we first seriously discussed the `Context`.
> > The bottom line right now is that there is a place in the `Context` for the `Connector`, and the code design and implementation is _optimized for_ and _takes advantage of_ that arrangement. If we choose to stay with this and use it as designed, we are basically done.
> 
> Is it an idea to separate the password bug and this design question from this PR and put it into a different PR? It would also be nice to have some instructions how to reproduce the password bug. We can then decide in our next meeting which road we want to take to resolve this.

Which password bug are you referring to?  I adjusted the password handling in `iBridges.py` because the given password was incorrectly retained in the GUI field between "Close session"s.  This was more an aesthetic issue than a bug.  Instead of opening an AUTH session, a FULL session was authenticated slowing things down a bit.

The password issue you refer to above:

> For some reason the iBridgesCli.py fails when one wants to login with a password and not with a cached password.

is simply an artifact of using the `Context` in a manner inconsistent with its design and is very easily remedied, as I have already shown.  If this solution is not acceptable, I can see what can be done on the `Context` side.  In any case, _some minor update_ will be required to the `iBridgesCli` logic.  I cannot implement the `Context` around two different use-cases without _some_ modification of both.

jpmcfarland:
> However, when I look at the session class, it still has a local variable for the config. I had expected to see references to the context singleton here instead: https://github.com/chStaiger/iBridges-Gui/blob/99c2a5776f3f3413ca8dbfcb83eaca20d827f941/irodsConnector/session.py#L19C1-L21C29

That was the original implementation, but it caused a circular reference: `Session().context.irods_connector.session`.  I believed it was not a serious issue, but we removed it nonetheless.  Now the configs are effectively _injected_, particularly the iRODS environment, because it unknown upon instantiation.

> 
> Now the iRODSconnector is saved in the config I am not sure if this will cause circular imports?

I don't understand what you mean by "iRODSconnector is saved in the config".

wurDevTim:
Apologies for the late reply, I wanted trace the development of the context class before commenting. It took some trials to get there with `git blame`. You also rightfully point out that I mixed up  `context` and `config`, in my mind they did essentially the same. 

Looking back it seems that the `iRODSconnector` has been part of the context since the start. 
I am not sure this PR is really the place for this design discussion but since it already started, the whole idea of the context singleton was to have one place for the configs. By _injecting_ the configs in the irods_client your again creating at least two, effectively 1+ the number of connector instance, copies which can all go out of sync. 

Is it not wiser to limit the scope of the context to the configs? and pass this singleton around through the program to ensure a single source of truth is used by all instances of the irodsconnector? 
This would still mean the CLI needs an instance of the context class to work, but to me it seems clearer and would enable us to create multiple irods connectors again.

chStaiger:
Just to summarise:

1. The Cli uses the Context to load and store the configs already. It just did not use the irodsConnector inside of the context object.
2. Currently one can put the irodsConnector back into the context but it was not mandatory for the irodsConnector instance to function. This would be changed by this PR. Before we do that we all need to be aware of that and agree upon that.
3. If I understand Tim correctly, he would prefer the Context being a class variable in the irodsConnector/irodsConnector.session and by that ensuring that all instances of the irodsConnector are dealing with the same parameters?!
4. I am a bit puzzled by the conundrum of this PR: if I instantiate an irodsConnector from information of the context and want to authenticate with a password, the irods environment parameters are not passed to the session object and hence the irodsConnector will fail to authenticate. However, when I instantiate the irodsConnector in exactly the same way but as a member of the context object the authentication just works fine. I would prefer investigating why that is rather than accepting a quick fix without understanding why it solves the issue.

chStaiger (approved):

PR #97, opened 2023-05-14
[96] Fix print to logging bugs

jpmcfarland:
Not only fix the bugs caused by the conversion, but make _all_ logging calls uniform in format.

~~It would be nice to (re)consolidate the two versions of `init_logger()`,  but that would require significant updates to `iBridgesCli.py` and is out of this scope of this issue.~~

Consolidating the `init_logger()` function is the subject of PR #102.

### DIFF
--- a/gui/IrodsBrowser.py
+++ b/gui/IrodsBrowser.py
@@ -351,8 +351,8 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget,
             else:
                 self.collTable.setRowCount(0)
                 self.errorLabel.setText("Collection does not exist.")
-        except irods.exception.NetworkException:
-            logging.exception('Something went wrong')
+        except irods.exception.NetworkException as error:
+            logging.error('Something went wrong: %r', error)
             self.errorLabel.setText(
                     "iRODS NETWORK ERROR: No Connection, please check network")
 
@@ -392,9 +392,9 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget,
             self._fill_metadata_tab(obj_path)
             self._fill_acls_tab(obj_path)
             self._fill_replicas_tab(obj_path)
-        except Exception as e:
+        except Exception as error:
             logging.error('Browser', exc_info=True)
-            self.errorLabel.setText(repr(e))
+            self.errorLabel.setText(repr(error))
 
     def loadSelection(self):
         # loads selection from main table into delete tab
--- a/gui/IrodsBrowser.py
+++ b/gui/IrodsBrowser.py
@@ -352,7 +352,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget,
                 self.collTable.setRowCount(0)
                 self.errorLabel.setText("Collection does not exist.")
         except irods.exception.NetworkException:
-            logging.exception("Something went wrong")
+            logging.exception('Something went wrong')
             self.errorLabel.setText(
                     "iRODS NETWORK ERROR: No Connection, please check network")
 
@@ -393,7 +393,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget,
             self._fill_acls_tab(obj_path)
             self._fill_replicas_tab(obj_path)
         except Exception as e:
-            logging.error('Browser',exc_info=True)
+            logging.error('Browser', exc_info=True)
             self.errorLabel.setText(repr(e))
 
     def loadSelection(self):
@@ -476,7 +476,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget,
                 self.errorLabel.setText(
                     "iRODS NETWORK ERROR: No Connection, please check network")
             except Exception as error:
-                logging.error("Upload :", fileSelect[0], "failed; \n\t", repr(error))
+                logging.error('Upload failed %s: %r', fileSelect[0], error)
                 self.errorLabel.setText(repr(error))
 
     def fileDownload(self):
@@ -505,7 +505,7 @@ class IrodsBrowser(PyQt6.QtWidgets.QWidget,
                 self.errorLabel.setText(
                     "iRODS NETWORK ERROR: No Connection, please check network")
             except Exception as error:
-                logging.error("Download :", parent+'/'+objName, "failed; \n\t", repr(error))
+                logging.error('Download failed %s/%s: %r', parent, objName, error)
                 self.errorLabel.setText(repr(error))
 
     # @PyQt6.QtCore.pyqtSlot(PyQt6.QtCore.QModelIndex)
--- a/gui/continousUpload.py
+++ b/gui/continousUpload.py
@@ -59,7 +59,7 @@ class contUpload(Thread, utils.context.ContextContainer):
                         folder_wfiles = self.tosync_dictionary.pop(filepath)
                         self.conn.upload_data(filepath, self.destColl, None, None, force=self.force)
                     else:
-                        logging.debug("Somethings going wrong. data folder in {filepath} not tracked")
+                        logging.debug("Something's going wrong. data folder in %s not tracked", filepath)
                 else: # Add files with folder as key
                     pathparts =  filepath.rsplit(path.sep, 1)
                     if pathparts[1] == "Data":
@@ -68,9 +68,9 @@ class contUpload(Thread, utils.context.ContextContainer):
                         else:
                             self.tosync_dictionary[pathparts[0]] = [pathparts[1] + path.sep + filename]
                     else:
-                        logging.debug("TODO, this should not happen? {filepath}  {filename}")
+                        logging.debug('TODO, this should not happen? %s  %s', filepath, filename)
             elif self.upload_mode == "f500":
-                logging.debug("TODO figure out how to do the F500 upload")
+                logging.debug('TODO figure out how to do the F500 upload')
 
             else: # "all"
                 self.conn.upload_data(new_file, self.destColl, None, None, force=self.force)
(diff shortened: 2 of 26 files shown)


jpmcfarland:
@wurDevTim @maartenschermer I know Christine is out for the next two weeks, but I want her to be aware.  I also would like reviews in the form of "none of the logging statements cause an error".  Everything looks fine to me, but I simply don't have the means to test everything.

wurDevTim (commented):
Heet John, 
Looked over the changes, I had a few small notes which I put inline. 
Also pressed around in the GUI, that looked good.

gui/continousUpload.py
```diff
@@ -59,7 +59,7 @@ def run(self):
                         folder_wfiles = self.tosync_dictionary.pop(filepath)
                         self.conn.upload_data(filepath, self.destColl, None, None, force=self.force)
                     else:
-                        logging.debug("Somethings going wrong. data folder in {filepath} not tracked")
+                        logging.debug("Something's going wrong. data folder in %s not tracked", filepath)
```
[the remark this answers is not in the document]
jpmcfarland: Stylistically, I use single quotes for all strings unless there is one or more within the string.  Escaping is possible, but not ideal, so I avoid it.  I don't even like `\`s to continue lines unless it is absolutely necessary.

gui/mainmenu.py
```diff
@@ -14,7 +14,7 @@
 
 class QPlainTextEditLogger(logging.Handler):
     def __init__(self, widget):
-        super(QPlainTextEditLogger, self).__init__()
+        super().__init__()
```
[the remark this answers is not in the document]
jpmcfarland: Yes, I see.  Thanks for the catch.

irodsConnector/session.py
```diff
@@ -170,19 +170,21 @@ def connect(self):
...
             else:
-                logging.error(f'{kw.RED}No iRODS session: "irods_env_file" not set!{kw.DEFAULT}')
+                logging.error(
+                    '%sNo iRODS session: "irods_env_file" not set!%s',
+                    kw.RED, kw.DEFAULT)
                 return
         options = {
             'irods_env_file': str(self.irods_env_file),
         }
-        logging.debug(f'{self.ienv=}')
```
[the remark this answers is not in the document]
jpmcfarland: Thanks!  I certainly did miss that one.

gui/mainmenu.py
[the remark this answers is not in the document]
jpmcfarland: Upon research, `QPlainTextEditLogger` is defined here.  So `super()` refers correctly to `logging.Handler`.

irodsConnector/session.py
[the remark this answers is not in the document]
jpmcfarland: No, I did remove that on purpose.  I want to use those statements only in select properties so I can see how often they are called.

gui/dataTransfer.py
```diff
@@ -252,7 +252,7 @@ def run(self):
                                                    self.coll.path, newPath, scope="checksum")
                 self.updLabels.emit(len(onlyIrods), len(diff))
         except:
-            logging.exception("dataTransfer.py: Error in getDataState")
+            logging.exception('dataTransfer.py: Error in getDataState')
```
maartenschermer, line 255: Not sure if `logging.exception` is a deliberate choice here (over `logging.error`), but I think we'd better stick to `logging.error`. `logging.exception` generates additional lines of stacktrace that are not useful here.

utils/context.py
jtersteegwur, line 1: Would it be possible to change the print on Line 264 to be logging as well?

https://github.com/chStaiger/iBridges-Gui/blob/17e7bffed8335dd88dd9a1de32e11a609b835622/utils/context.py#L264

gui/IrodsBrowser.py
```diff
@@ -393,7 +393,7 @@ def fillInfo(self, index):
             self._fill_acls_tab(obj_path)
             self._fill_replicas_tab(obj_path)
         except Exception as e:
```
maartenschermer, line 395: For concistency's sake, `e` should be `error`, I think.

iBridges.py
```diff
@@ -186,9 +186,9 @@ def login_function(self):
             self.setCursor(PyQt6.QtGui.QCursor(PyQt6.QtCore.Qt.CursorShape.ArrowCursor))
             return
         except Exception as unknown:
```
maartenschermer, line 188: For concistency's sake, `e` should be `error`, I think.

iBridgesCli.py
```diff
@@ -242,11 +242,10 @@ def connect_irods(cls, irods_env):
 
                 break
             except AssertionError as exception:
```
maartenschermer, line 244: For concistency's sake, `exception` should be `error`, I think.

utils/elab_plugin.py
```diff
@@ -34,8 +34,9 @@ def setup(self, calling_class):
             try:
                 self.elab.updateMetadataUrl(group=config['group'], experiment=config['experiment'])
             except ValueError as exception:
-                logging.error("ELN groupID %s or experimentID %s not set or invalid. (%s)",
```
maartenschermer, line 37: For consistency's sake, I think `exception` should be `error` (occurs several more times below).

jtersteegwur:
Along similar lines to @maartenschermer 's comments:
https://github.com/chStaiger/iBridges-Gui/blob/jpm/bugfix/96_print_to_logging/gui/irodsTreeView.py#L55

Could/should this be an error?

irodsConnector/dataOperations.py
```diff
@@ -321,7 +321,7 @@ def upload_data(self, source: str, destination: irods.collection.iRODSCollection
         if not force:
             space = self.resc_man.resource_space(res_name)
             if size > (space - buff):
-                logging.info(
+                logging.error(
```
maartenschermer, line 324: I think putting the actual word 'ERROR' in the error string is superfluous (several more instances below).

irodsConnector/meta.py
```diff
@@ -28,9 +28,9 @@ def add(self, items: list, key: str, value: str, units: str = None):
             try:
                 item.metadata.add(key.upper(), value, units)
             except irods.exception.CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME:
-                logging.error(kw.RED+"INFO ADD META: Metadata already present"+kw.DEFAULT)
+                logging.error('%sADD META: Metadata already present%s', kw.RED, kw.DEFAULT)
```
maartenschermer, line 31: I think we should remove the colour codes (or at leat move somehow them to the logging handler).

irodsConnector/permission.py
```diff
@@ -106,11 +106,10 @@ def set_permissions(self, perm: str, path: str, user: str = '',
                     self.data_man.collection_exists(path):
                 self.sess_man.irods_session.permissions.set(acl, recursive=recursive, admin=admin)
         except irods.exception.CAT_INVALID_USER as ciu:
```
maartenschermer, line 108: I think for consistency, `ciu` and `cia` should become `error`.

utils/context.py
jpmcfarland: Yep!

iBridges.py
jpmcfarland: Consistency is nice.

utils/elab_plugin.py
jpmcfarland: Got it.

irodsConnector/tickets.py
```diff
@@ -52,7 +52,7 @@ def create_ticket(self, obj_path: str, expiry_string: str = '') -> tuple:
             try:
                 expiration_set = self._modify_ticket(ticket, expiry_string)
             except Exception as error:
-                logging.info('Could not set expiration date: %s', str(error))
+                logging.info('Could not set expiration date: %r', error)
```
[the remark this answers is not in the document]
jpmcfarland: Absolutely.

irodsConnector/dataOperations.py
jpmcfarland: All updated.

irodsConnector/permission.py
jpmcfarland: Makes sense as long as they are only used locally and not needed again.

irodsConnector/meta.py
jpmcfarland: As far as I can tell, there is no per-type formatting, only one format per handler.  I think the best I can do is create a set of wrapper functions (or even a single function) to call the logger methods with the appropriate ASCII color codes surrounding the logging message.  Perhaps something like:

```python
def log(log_method, log_message):
    mdict = {
        logging.debug: '%s',
        logging.info: 'kw.YELLOW%skw.DEFAULT',
        logging.warning: 'kw.ORANGE%skw.DEFAULT',
        logging.error: 'kw.RED%skw.DEFAULT',
    }
    log_method(mdict[log_method], log_message)
```

Unfortunately, every single logging statement will have to be updated and the advantage of optional evaluation will be lost.

jpmcfarland: I think a simple string converting function is the best solution.  A little more work when writing the logging statement, but retains all the other advantages.

jpmcfarland: Found the `correct` solution: **add custom formatting targets to the record factory.**

jpmcfarland: See: https://github.com/chStaiger/iBridges-Gui/pull/97/commits/0098e934972b4916f2d29260b69714ca2c6d2a55

jpmcfarland:
@jtersteegwur @maartenschermer @wurDevTim Please have another quick sanity check.

Approve or leave a message so I know you have seen it.  Thanks!

irodsConnector/meta.py
maartenschermer: Looks good! Is it possible to put the `init_logger()` somewhere more central, maybe in utils, so `iBridgesCli.py` can use it too?

maartenschermer (approved):

Issue #94, opened 2023-05-09
JsonConfig does not reload the json when file path is changed


chStaiger:

chStaiger:
Fixed in https://github.com/chStaiger/iBridges-Gui/commit/a3edc26954b7d3f25f388bbf95423dc453d8844c

Issue #96, opened 2023-05-12
Fix bugs converting from print statements


jpmcfarland:
Some malformed logging statements were caused by blindly converting from the original print statements.  The main issue is from statements such as:

```python
print(a, b, c)
```

converted to:

```python
logging.info(a, b, c)
```

causing an error because arguments after the first (the message) are considered to be string formatting arguments to the first.

chStaiger:
I merged PR #97. Can this issue be closed?

jpmcfarland:
Of course!

Issue #82, opened 2023-04-07
Code style cleanup


jpmcfarland:
_This issue has a broad scope and we will need to decide when it should be done and discuss how it should be done.  Sooner is better than later.  And needless to say, any new code should certainly follow a convention.  I have implicitly proposed PEP 8 as it is the de facto standard for general Python code formatting, and similarly NumPy docstrings._

Consistently apply the PEP 8 conventions across the board in all manually written Python code and filenames/directories following:

https://peps.python.org/pep-0008/

Certain external library naming conventions cannot be changed, namely PyQt, but those that can, the entities in the UI files that intersect with the code (e.g., target widget names), should adhere where possible.

I can propose names of top-level modules, classes, and similar so we can discuss them prior to implementation.

It may also be an idea to pick a convention for the directory structure that is more typical of open-source Python projects, but still allows the various distribution, testing, and other tools to function properly.  For instance, something like this may be ideal:

- iBridges (repository name)
  - doc/
  - ibridges/
    - connector/
    - gui/
    - test/
    - utils/
    - ibridges.py
  - install.py
  - LICENSE
  - README.md
  - requirements.txt

As for docstrings, **ALWAYS USE DOCSTRINGS**.  The NumPy convention for docstrings is quite clear and readable:

https://numpydoc.readthedocs.io/en/latest/format.html

Current tasks:

- update internal naming (class/method/function/variable)
- [x] iBridges.py
- [ ] iBridgesCli.py
- [ ] gui/*py
- [ ] gui/ui_files/*
- [ ] irodsConnector/*py
- [ ] utils/*py

Other tasks:

- update external (file/directory) naming?
- restructure modules?

jpmcfarland:
Now that PR #80 is merged, I can start adding missing documentation and type hinting, and basic internal naming updates I have been working on.  There will inevitably be some minor code optimizations and bug fixes as I also update unit tests.

It will be good to _eventually_ discuss the more major changes such as file remaining and repository restructuring, particularly the arguments for and against such an effort.

chStaiger:
The package will get a complete make over ... closed for now

PR #98, opened 2023-05-14
Cli integration

### LINKED ISSUE #93: irods_env_file not set in session.py
The session.py only reads the last irods_env from the ibridges_conf, but when we want to set a new irods_env_file it gets ignored.

chStaiger:
Some bugs that occurred after last merge:
- Session was only created from ibridges_config['last_ienv'] --> Fixed for GUI
- Search function does not yield results any longer

TODO:

- [ ] Fix login (irods session creation) for GUI
- [ ] Fix login (irods session creation) for CLI 
- [ ] Fix and test search 

This PR also fixes  #93  and #94


jpmcfarland:
I believe you intended to merge into `develop` not `master`.

chStaiger:
> I believe you intended to merge into `develop` not `master`.

Yes, changed. Tomorrow I will tick off the other two items on the list and then it’s ready for review.

chStaiger:
I changed:

- Autorelaod of irods environment when the irods ate is changed in context (affects also GUI)
- Instantiation of meta in manager.py (affects also GUI)
- Some small fixes in Cli-python

iBridges.py
```diff
@@ -38,15 +38,15 @@
...
+                       gui.ui_files.irodsLogin.Ui_irodsLogin):
     """Definition and initialization of the iRODS login window.
 
     """
     icommands = False
     this_application = ''
+    context = utils.context.Context()
 
-    def __init__(self):
+    def __init__(self, context):
```
jpmcfarland, line 49: The passed `context` argument is not used.  Either choose to pass it and set it in the constructor as a class attribute or set it as an initial class attribute as above (L47).

irodsConnector/session.py
```diff
@@ -171,14 +171,6 @@ def connect(self):
 
         """
         logging.debug(f'{self.irods_env_file=}')
-        if not self.irods_env_file:
```
jpmcfarland, line 174: This works fine.  Is it simply undesired functionality?

utils/json_config.py
```diff
@@ -57,7 +57,8 @@ def config(self) -> dict:
             exists.  The empty persistent dictionary otherwise.
 
         """
-        if self._config == {}:
+        if self._config == {} or self.reload:
```
jpmcfarland, line 60: `self.reload` is unnecessary.  When the `self.filepath` is changed, it is a new configuration.  The correct procedure is to first `reset()` the configuration prior to setting the new filepath:

```python
config.reset()
config.filepath = new_filepath
```

and the configuration will reload.  This is not done automatically or the current configuration will be lost.  If the previous configuration it to bne retained, it should be saved prior to the `reset()`.

This should already be done in the GUI.  I don't know about the CLI.

iBridges.py
chStaiger: That's still an artefact, it should not be passed as a parameter.

irodsConnector/session.py
chStaiger: I know and I like the feature, I just think that the frontend (GUI or Cli) which check all parameters before giving them to the backend should decide which irods_environment to use, the previous one or if the user says so, the new one. The backend should not "correct" the users' choices, that should be up to the frontend.

utils/json_config.py
chStaiger: I will try it out in the Cli!

chStaiger:
There is still a small problem with the Cli. 
The previous version would ignore the irodsenvpath from the config file:
We set a default ($HOME/.irods/irods_environment.json) to irods_env when the client starts from arguments but since that is then not None, the value from the config will never be used in [`self.irods_env =  irods_env or self.get_config('iRODS', 'irodsenv') or self._clean_exit("need iRODS environment file", True)`](https://github.com/chStaiger/iBridges-Gui/blob/93b1408feb34f397e752c4b63c985d8205942123/iBridgesCli.py#L85)

Now I changed the two arguments around the *or*, with the effect, that we cannot overwrite config parameters from command line.

irodsConnector/session.py
jpmcfarland: Perhaps it is an idea to add an option to the configuration to allow a user to choose this behavior?  Something like a boolean value to `"use_prev_ienv": False`.

I like the idea of `"last_ienv": ""` to indicate this, but the setting of that is automatic and I don't see that changing.  Also, given that "last_ienv" gives the default environment file selection in the GUI, a new mechanism for that would be needed.

iBridgesCli.py
```diff
@@ -71,6 +71,9 @@ def __init__(self,                      # pylint: disable=too-many-arguments
         self.config_file = None
         self.download_finished = None
         self.upload_finished = None
+        self.irods_conn = None
+
+        default_irods_env = os.path.join(str(os.getenv('HOME')), '.irods', 'irods_environment.json')
```
jpmcfarland, line 76: You can use a `LocalPath` here:

```python
default_irods_env = utils.path.LocalPath(utils.context.IRODS_DIR, 'irods_environment.json').expanduser()
```

When assigning to a `JsonConfig.filepath`, `expanduser()` is called automatically.

I didn't have a need for it previously, but you could also add `DEFAULT_IRODS_ENV_FILE` to `context.py`:

```python
DEFAULT_IRODS_ENV_FILE = f'{IRODS_DIR}/irods_environment.json'
```

and thus it becomes:

```python
default_irods_env = utils.path.LocalPath(utils.context.DEFAULT_IRODS_ENV_FILE).expanduser()
```

You can update L178-179 to use these constants as well.

irodsConnector/manager.py
```diff
@@ -113,9 +109,7 @@ def irods_env_file(self, filepath: str):
 
         """
         self._irods_env_file = filepath
-        logging.debug(f'setting: {self._irods_env_file=}')
-        if self.session:
-            self.session.irods_env_file = filepath
+        logging.info(f'setting: {self._irods_env_file=}')
```
jpmcfarland, line 112: Unless you need to know the value of the variable, it is better to indicate only that the variable is set.  If you need to know the value, using `logging.debug` with %-formatting will be more efficient when properties like this are called frequently:

```python
logging.debug('setting: %s', self._irods_env_file)
```

See PR #97 for more examples.  I can also update this there while I'm doing the rest.

utils/context.py
```diff
@@ -90,7 +90,7 @@ def ibridges_configuration(self) -> json_config.JsonConfig:
             Configuration instance.
 
         """
-        if self._ibridges_configuration is None:
+        if self._ibridges_configuration is None or self._ibridges_configuration.config is {}:
```
jpmcfarland, line 93: The config file loading is intended only to be called the one time, when the `_ibridges_configuration` has never been set.  If it is re-set like this, a new object will be created and previous pointers to the old object will need to be updated to this new object _everywhere_ it is used.  This is very undesirable!

In the context of Issue #82, I have this alternate implemention that should work better:

```python
        if self._ibridges_configuration is None:
            if not self.ibridges_conf_file:
                self.ibridges_conf_file = DEFAULT_IBRIDGES_CONF_FILE
            filepath = path.LocalPath(self.ibridges_conf_file).expanduser()
            if not filepath.parent.is_dir():
                filepath.parent.mkdir()
            if not filepath.is_file():
                filepath.write_text(json.dumps(IBRIDGES_CONF_TEMPLATE))
            self._ibridges_configuration = json_config.JSONConfig(filepath)
        elif self.ibridges_conf_file != self._ibridges_configuration.filepath:
            self._ibridges_configuration.reset()
            self._ibridges_configuration.filepath = self.ibridges_conf_file
```

The new `filepath` and empty `config` dictionary will then cause the `JsonConfig` to load the configuration from the new file within the original instance.

Incidentally, this resolves an esoteric bug where the iBridges configuration is wiped and recreated when connecting after a 'Close session'.

utils/context.py
```diff
@@ -188,9 +185,9 @@ def irods_environment(self) -> json_config.JsonConfig:
             Configuration instance.
 
         """
-        if self._irods_environment is None:
-            # TODO add existence check, running "iinit" when missing?
+        if self._irods_environment is None or self._irods_environment.config is {}:
```
jpmcfarland, line 188: Same comment as for L93 with this alternative code:

```python
        if self._irods_environment is None:
            # TODO add existence check, running "iinit" when missing?
            self._irods_environment = json_config.JSONConfig(self.irods_env_file)
        elif self.irods_env_file != self._irods_environment.filepath:
            self._irods_environment.reset()
            self._irods_environment.filepath = self.irods_env_file
        return self._irods_environment
```

utils/json_config.py
```diff
@@ -58,6 +58,7 @@ def config(self) -> dict:
 
         """
         if self._config == {}:
+            self.reload = False
```
jpmcfarland, line 61: You missed removing this.

iBridgesCli.py
jpmcfarland: Perhaps it is even better to use this assignment:

```python
DEFAULT_IBRIDGES_CONF_FILE = path.LocalPath(
    IBRIDGES_DIR, 'ibridges_config.json').expanduser()
DEFAULT_IRODS_ENV_FILE = path.LocalPath(
    IRODS_DIR, 'irods_environment.json').expanduser()
```

Then:

```python
default_irods_env = utils.context.DEFAULT_IRODS_ENV_FILE
```

is possible.

iBridges.py
```diff
@@ -147,20 +148,23 @@ def login_function(self):
...
         self.context.save_ibridges_configuration()
         password = self.passwordField.text()
-        self.conn.password = password
+        self.context.irods_connector.password = password
         logging.debug(f'IRODS PASSWORD SET: {"*"*len(password)*2}')
         try:
-            self.conn.connect()
+            self.context.irods_connector.irods_env_file = self.context.irods_env_file
+            self.context.irods_connector.irods_environment = self.context.irods_environment
+            self.context.irods_connector.ibridges_configuration = self.context.ibridges_configuration
```
maartenschermer, line 166: These three lines (164, 165, 166) don't really need to be within the `try...except`, I guess.

iBridgesCli.py
maartenschermer, line 76: Not sure `os.getenv('HOME')` makes sense in a Windows environment. Maybe John's suggestion also takes care of that?

chStaiger:
Just quickly checked compatibility between this PR and PR #96. Changes are only minor. It hardly makes any difference what is merged first.

irodsConnector/manager.py
jpmcfarland: This and L138 really should be _debug_ level.  The user normally does not need this information unless something unusual is happening.  That is when _debug_ is used.

Issue #91, opened 2023-05-09
move logging initialisation to utils


chStaiger:
The init of the logging is in iBridges.py. But there it cannot be accessed by the CLI --> move it to utils.utils.py

chStaiger:
Done. the Cli makes use of the logging initialisation in the utils. We can also switch to that in the Gui (iBridges.py).

chStaiger:
Taken up by issue #102 and PR #102 --> close

jpmcfarland:
> Taken up by issue https://github.com/chStaiger/iBridges-Gui/pull/102 and PR https://github.com/chStaiger/iBridges-Gui/pull/102 --> close

But this _is_ the Issue for PR #102.

It's fine to close now, but that usually happens when the associated PR is merged.

PR #102, opened 2023-05-19
[91] Unify logging initialization

jpmcfarland:
I would like to unify the use of the logger initialization.  The tricky part is to initialize logging with the default logging level as soon as possible and setting the user's preferred logging level _after_ the **Context** is initialized.  Doing that at the same time is likely to cause issues, one of which is initializing the context _before_ intended.

As of the commit associated with Issue #91, there are two copies of the **init_logger()** function.  I have updated this a bit in PR #97 by moving **set_log_level()** to `utils/utils.py`, but I would like to merge the functionality of **init_logger()** as well so there is only one copy.  This PR updates `iBridges.py` and `iBridgesCli.py` to use the same functions, but in slightly different ways.

The only conceptual problem I'm having in `iBridgesCli.py` is that **Context** is formally meant to be initialized in **connect_irods()**, but the logger is initialized in the constructor.  Setting of the log level from the configuration was first done via **utils.utils.init_logger()**, but now via **utils.utils.set_log_level()**.  This is done to meet the condition mentioned at the top.

Questions:

1. Why is **connect_irods()** a class method?  It is not used anywhere else I can see except as a normal method?
2. Can the **Context** initialization be moved up to the constructor?
3. Better yet, could perhaps a **main()** function can be used for argument parsing/running and the logging/**Context** be initialized there instead?  That would be more in line with how `iBridges.py` is doing it.

### DIFF
--- a/iBridges.py
+++ b/iBridges.py
@@ -224,60 +224,19 @@ def closeClean():
         context.irods_connector.cleanup()
 
 
-def init_logger():
-    """Initialize the application logging service.
-
-    """
-    logger = logging.getLogger()
-    logdir = utils.path.LocalPath(utils.context.IBRIDGES_DIR).expanduser()
-    logfile = logdir.joinpath(f'{THIS_APPLICATION}.log')
-    log_formatter = logging.Formatter(
-        '[%(asctime)s] {%(filename)s:%(lineno)d} %(levelname)s - %(message)s')
-    file_handler = logging.handlers.RotatingFileHandler(logfile, 'a', 100000, 1)
-    file_handler.setFormatter(log_formatter)
-    logger.addHandler(file_handler)
-    stream_handler = logging.StreamHandler(sys.stdout)
-    stream_handler.setFormatter(log_formatter)
-    logger.addHandler(stream_handler)
-    # Indicate start of a new session
-    with open(logfile, 'a', encoding='utf-8') as logfd:
-        logfd.write('\n\n')
-        underscores = f'{"_" * 50}\n'
-        logfd.write(underscores * 2)
-        logfd.write(f'\t\t{datetime.datetime.now().isoformat()}\n')
-        logfd.write(underscores * 2)
-
-
-def set_log_level(log_level: int):
-    """Set the log level excluding DEBUG-level entries from other
-    modules.
-
-    Parameters
-    ----------
-    log_level : int
-        Level to set the current logger.
-
-    """
-    logging.getLogger().setLevel(log_level)
-    if log_level == logging.DEBUG:
-        for logger in logging.Logger.manager.loggerDict.values():
-            if hasattr(logger, 'name') and logger.name != 'root':
-                logger.disabled = True
-
-
 def main():
     """Main function
 
     """
     # Initialize logger first because Context may want to log as well.
-    init_logger()
+    logdir = utils.path.LocalPath(utils.context.IBRIDGES_DIR).expanduser()
+    utils.utils.init_logger(logdir, THIS_APPLICATION)
     # Singleton Context
     context = utils.context.Context()
     context.application_name = THIS_APPLICATION
     # Context is required to get the log_level from the configuration.
-    verbose = context.ibridges_configuration.config.get('verbose', 'info')
-    log_level = LOG_LEVEL.get(verbose, logging.INFO)
-    set_log_level(log_level)
+    # Here, it is taken from the configuration if not specified.
+    utils.utils.set_log_level()
     context.irods_connector = irodsConnector.manager.IrodsConnector()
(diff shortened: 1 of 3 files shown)


chStaiger:
> As of the commit associated with Issue #91, there are two copies of the **init_logger()** function. I have updated this a bit in PR #97 by moving **set_log_level()** to `utils/utils.py`, but I would like to merge the functionality of **init_logger()** as well so there is only one copy. This PR updates `iBridges.py` and `iBridgesCli.py` to use the same functions, but in slightly different ways.
> 

I think the way it is now with init_logger and set_log_level in utils/utils.py is the right place. Default loglevel 'debug' is also what is most common when no other options are given or known.

> The only conceptual problem I'm having in `iBridgesCli.py` is that **Context** is formally meant to be initialized in **connect_irods()**, but the logger is initialized in the constructor. Setting of the log level from the configuration was first done via **utils.utils.init_logger()**, but now via **utils.utils.set_log_level()**. This is done to meet the condition mentioned at the top.
> 

> Questions:
> 
> 1. Why is **connect_irods()** a class method?  It is not used anywhere else I can see except as a normal method?
That is a question for @maartenschermer 

> 2. Can the **Context** initialization be moved up to the constructor?
Yes, will do so in the PR #98 

> 3. Better yet, could perhaps a **main()** function can be used for argument parsing/running and the logging/**Context** be initialized there instead?  That would be more in line with how `iBridges.py` is doing it.

I think that would be in the current code `IBridgesCli.from_arguments`. The CLI is written such that one can overwrite arguments in the config by passing them in the command-line call. That is done there and everything that is not known from the arguments is set in the constructor. So I would initialise the Context in the constructor, fill the unknown arguments either from the iBridges and iRODS configs. @maartenschermer Do you agree?

jpmcfarland:
> I think the way it is now with init_logger and set_log_level in utils/utils.py is the right place. Default loglevel 'debug' is also what is most common when no other options are given or known.

Actually, _debug_ is only used for, ahem, debugging purposes.

Try to remove the silencing code in `set_log_level()` when running iBridges with _debug_ level set and you'll see what I mean.  Before you even get to the login dialog, you will be assaulted with nearly one hundred lines of log entries, then many many hundreds after you press connect as thee panels load.  These are from dependent libraries (that's why they are hidden), but it gives you an idea what other developers consider _debug_ level messages.

I think the default really should be left at _info_.  If there are iBridges entries currently at _debug_ level that you feel a user needs to see on a regular basis, please consider changing those to _info_ instead.

chStaiger:
Sorry, I meant 'info'.

maartenschermer:
> I think that would be in the current code `IBridgesCli.from_arguments`. The CLI is written such that one can overwrite arguments in the config by passing them in the command-line call. That is done there and everything that is not known from the arguments is set in the constructor. So I would initialise the Context in the constructor, fill the unknown arguments either from the iBridges and iRODS configs. @maartenschermer Do you agree?

Yes, I agree. The idea is to be able to use different configs by passing the environment file to use from the command line, and have that override any defaults. I was planning to integrate this behaviour with the Context-class, but was waiting for that to be stable.

jpmcfarland:
> > 3. Better yet, could perhaps a **main()** function can be used for argument parsing/running and the logging/**Context** be initialized there instead?  That would be more in line with how `iBridges.py` is doing it.
> 
> I think that would be in the current code `IBridgesCli.from_arguments`. The CLI is written such that one can overwrite arguments in the config by passing them in the command-line call. That is done there and everything that is not known from the arguments is set in the constructor. So I would initialise the Context in the constructor, fill the unknown arguments either from the iBridges and iRODS configs. @maartenschermer Do you agree?

It certainly works fine the way it is, but I think it would be more clear and simpler to develop if it followed a more typical structure.  Such structures evolved from experience with reason.  The typical start-up script is structured something like:

```python
class MyProgram:

    # a bunch of code


def parse_arguments(*args) -> tuple:

    # a bunch of argparse definitions


def main() -> int:

    args = parse_arguments(sys.argv[1:])
    prog = MyProgram(*args)
    return prog.run()


if __name__ == '__main__':
    exit(main())
````

The main program class is responsible for the program, and support function(s) are responsible for translating the user input and providing it to the program class.

This is just a recommendation and food for thought.

jpmcfarland:
> Sorry, I meant 'info'.

No worries.  Have a look at how much debugging info iRODS has using the change in this commit: https://github.com/chStaiger/iBridges-Gui/pull/105/commits/d8460e3639381a1de5a12d0cf2e62aa26fc27359.  I accidentally committed this to a different PR thinking it was this one.

This iRODS debug info is quite interesting and likely will be very helpful in the future.

Issue #103, opened 2023-05-24
Connectors issues


jpmcfarland:
While surveying logging updates, I came across a potential set of issues related to the connectors.  The `manager.IrodsConnector` and the `Icommands.IrodsConnectorIcommands` classes, though having parallel names and functionality are not completely parallel.  Indeed, the iCommands connector is actually contained within the "main" connector.  This is by design, but the latter expects an `irods_put` and `irods_get` method that is missing and some type hints need adjusting.  Also, the `tickets.Tickets` class attempts switching between iCommands/Python functionality.  This should actually be in the main connector where the determination of the existence of the iCommands can be determined.

I believe it may be that some functionality has been adjusted and not all references updated appropriately.  This needs to be looked at to see if there may be more issues and to get things fixed.

I'm happy to have a look unless somebody else prefers to do so.

jpmcfarland:
In addition to these issues, I also discovered that regardless of the setting of `Up-/Download with iCommands (linux only)`, if a Linux system has the iCommands present, it will use them and _not_ the Python versions.  This is likely another oversight.

jpmcfarland:
Upon a more detailed look, I removed the iCommands functionality as a concept and for testing.  The extra method did not appear necessary.  If it desired, it will have to be written into the iCommands connector regardless.

wurDevTim:
> In addition to these issues, I also discovered that regardless of the setting of `Up-/Download with iCommands (linux only)`, if a Linux system has the iCommands present, it will use them and _not_ the Python versions. This is likely another oversight.
> 
> This was discovered by noticing there were no logging messages for transfers where there should have been. Logging for transfers with iCommands will need to be update as well.

If I recall correctly we intend to remove the icommands up/download completely as soon as Jos finishes the new sync code. 
The reason for this is that besides the bug in the Tickets in the python API there is nothing the icommands can do that the pythonAPI cannot.

chStaiger:
> > In addition to these issues, I also discovered that regardless of the setting of `Up-/Download with iCommands (linux only)`, if a Linux system has the iCommands present, it will use them and _not_ the Python versions. This is likely another oversight.
> > This was discovered by noticing there were no logging messages for transfers where there should have been. Logging for transfers with iCommands will need to be update as well.
> 
[...]

PR #105, opened 2023-05-25
[103] Connectors issues

jpmcfarland:
All the problems I found related to Issue #103 appear to be fixed by this PR.

### DIFF
--- a/iBridges.py
+++ b/iBridges.py
@@ -43,7 +43,7 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog,
     """Definition and initialization of the iRODS login window.
 
     """
-    icommands = False
+    use_icommands = None
     this_application = ''
 
     def __init__(self):
@@ -110,20 +110,21 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog,
         """
         if self.standardButton.isChecked():
             self._init_envbox()
-            self.icommands = False
+            self.use_icommands = False
 
     def setup_icommands(self):
         """Check the state of the radio button for using iCommands.
-        This includes a check for the existance of the iCommands on the
+        This includes a check for the existence of the iCommands on the
         current system.
 
         """
         if self.selectIcommandsButton.isChecked():
             self.icommandsError.setText('')
-            if self.conn.icommands:
-                self.icommands = True
+            if self.conn.icommands.has_icommands:
+                self.use_icommands = True
                 # TODO support arbitrary iRODS environment file for iCommands
             else:
+                self.use_icommands = False
                 self.icommandsError.setText('ERROR: no iCommands found')
                 self.standardButton.setChecked(True)
 
@@ -154,6 +155,7 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog,
             self.setCursor(PyQt6.QtGui.QCursor(PyQt6.QtCore.Qt.CursorShape.ArrowCursor))
             return
         self.setCursor(PyQt6.QtGui.QCursor(PyQt6.QtCore.Qt.CursorShape.WaitCursor))
+        self.conn.use_icommands = self.use_icommands
         self.conf['last_ienv'] = irods_env_file.name
         self.context.save_ibridges_configuration()
         password = self.passwordField.text()
--- a/iBridges.py
+++ b/iBridges.py
@@ -24,6 +24,7 @@ DEFAULT = '\x1b[0m'
 RED = '\x1b[1;31m'
 YELLOW = '\x1b[1;33m'
 LOG_LEVEL = {
+    'fulldebug': logging.DEBUG - 5,
     'debug': logging.DEBUG,
     'info': logging.INFO,
     'warn': logging.WARNING,
@@ -235,6 +236,8 @@ def init_logger():
     def new_factory(*args, **kwargs) -> logging.LogRecord:
         """Custom record factory"""
         record = old_factory(*args, **kwargs)
+        # Limit the size of the log message to something sane.
+        record.msg = record.msg[:1024]
         record.prefix = ''
         record.postfix = ''
         if record.levelname == 'WARNING':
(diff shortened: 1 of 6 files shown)


jpmcfarland:
The last commit was intended for PR #102 but got put into here instead.  Too many active PRs ;)

It's a small change that's not too out of place, so I don't worry.

iBridges.py
```diff
@@ -233,6 +236,8 @@ def init_logger():
     def new_factory(*args, **kwargs) -> logging.LogRecord:
         """Custom record factory"""
         record = old_factory(*args, **kwargs)
+        # Limit the size of the log message to something sane.
+        record.msg = record.msg[:1024]
```
wurDevTim, line 240: Should this be a define at the top of the file or can this constant stay hidden in the code?

irodsConnector/tickets.py
```diff
@@ -43,41 +43,13 @@ def create_ticket(self, obj_path: str, expiry_string: str = '') -> tuple:
...
-        ticket_id = ''.join(choice(ascii_letters) for _ in range(20))
-        ticket = irods.ticket.Ticket(self.sess_man.irods_session, ticket_id)
+        ticket = irods.ticket.Ticket(self.sess_man.irods_session)
         ticket.issue('read', obj_path)
         logging.info('CREATE TICKET: %s: %s', ticket.ticket, obj_path)
         expiration_set = False
         if expiry_string != '':
             try:
-                expiration_set = self._modify_ticket(ticket, expiry_string)
+                expiration_set = ticket.modify('expire', expiry_string) == ticket
```
wurDevTim, line 52: Do you know if this bug is fixed now?

iBridges.py
jpmcfarland: Hmm, the number of characters is arbitrary and meant only to prevent MBs or more of file contents to enter the log file during debug-level transfers.  It certainly can be a constant to make it clearer.  Also, this code is destined to be in the `utils` module shortly in any case.

irodsConnector/manager.py
```diff
@@ -260,20 +263,11 @@ def get_dataobject(self, path: str) -> irods.data_object.iRODSDataObject:
     def get_irods_size(self, path_names: list) -> int:
         return self.data_op.get_irods_size(path_names)
 
-    def has_icommands(self) -> bool:
```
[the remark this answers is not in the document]
jpmcfarland: Yeah, I puzzled over this as well.  I think the simplest answer is that the iCommands connector is currently used as a sort of sub-connector that will likely go away soon.  It seemed best not to abstract the current single call that is effectively deprecated.

irodsConnector/tickets.py
jpmcfarland: I have no easy way to test it fully.  I expect somebody else can.  The limit if my testing is sanity checks such as finding no syntax errors.  Do we have any kind of unit tests for these modules?

iBridges.py
jpmcfarland: Updated.

irodsConnector/tickets.py
chStaiger: That is true. With the python API one could not set an expiry date and one had no function to clean up old tickets (not as user and not as admin). So we would still need the icommands here.

trel: Is there a python-irodsclient issue noting this lack of functionality?   A quick search didn't reveal one...

/cc @d-w-moore

jpmcfarland: I have tested the PRC functionality to create and modify a Ticket with the expiration date using the same methods in the connector and they all work fine.  I have not seen any functionality in iBridges for clearing tickets, but the iRODS Ticket() class does provide a working `delete()` method.  Unfortunately, querying for tickets is a bit round-about, but it does work.  What other functionality is needed?

When I tried all this using the iBridges ticket login, I keep getting segmentation faults when I change focus under macOS.  I'm trying under Linux, but I'm having to solve what looks like minor bugs.

Shall we make adding PRC query and delete functionality to this Issue/PR or should we create a new one?  I any case, I would like to solve the bugs I have just seen before this PR is ready to merge.

chStaiger: I have just tested the functionality through the GUI on Mac. So the ticket was created with the Python API and an expiry date was successfully set. I confirmed that through the icommands. Issue is solved.

The deletion of tickets according to their creation date or expiry date, has not been implemented yet in our backend. I will pout it in the issue list. I can remember that there was an issue with table names and how the table was accessed through the c++ API?! @trel

d-w-moore: @jpmcfarland I also hadn't noted (yet) any such bugs with the PRC ticket functionality.  Tickets may also be queried and deleted (examples are in the tests), but I certainly wouldn't deny the possibility of unknown bugs and unexpected behaviors.

irodsConnector/manager.py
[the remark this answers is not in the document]
chStaiger: Just to confirm, in the CLI I used the manager.has_icommands.

iBridges.py
```diff
@@ -113,20 +115,21 @@ def setup_standard(self):
...
         current system.
 
         """
         if self.selectIcommandsButton.isChecked():
             self.icommandsError.setText('')
-            if self.conn.icommands:
-                self.icommands = True
+            if self.conn.icommands.has_icommands:
+                self.use_icommands = True
                 # TODO support arbitrary iRODS environment file for iCommands
```
chStaiger, line 130: The icommands do not work with an arbitrary irods_environment.json. It has to always be ~/.irods/irods_environment.json
So the hint for the GUI is: if the standard irods_environment.json is selected and the icommands are available, use the IrodsConnectorIcommands otherwise use the IrodsConnector.

irodsConnector/tickets.py
jpmcfarland: It turns out the only bugs I'm currently seeing are related to a module requiring updates after the recent set of major refactoring.  I believe this was simply a low priority task and I stumbled into it by noticing an unusual inconsistency.

iBridges.py
trel: Near the bottom of https://docs.irods.org/4.3.0/system_overview/configuration/#irodsirods_environmentjson

> To use an environment file other than ~/.irods/irods_environment.json, set IRODS_ENVIRONMENT_FILE to load from a different location:
> 
> `export IRODS_ENVIRONMENT_FILE=/full/path/to/different.json`
> 
> Other individual environment variables can be set by using the UPPERCASE versions of the variables named above, for example:
> 
> `export IRODS_CWD='/tempZone/home/public'`
>

chStaiger: I can add it to the CLI. So for the GUI, all is fine then.

irodsConnector/manager.py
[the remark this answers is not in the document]
chStaiger: Somehow the `has_icommands` vanished from my code ... I will adjust the CLI once this PR is done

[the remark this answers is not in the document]
jpmcfarland: Weird!  Oki doki.

Issue #108, opened 2023-05-30
Unify logging initialization


jpmcfarland:
Moving `init_logger()` out of `utils` led to a divergence of the logging initialization.  Lets fix that by unifying it back into `utils`.

chStaiger:
Merged! Close?

PR #109, opened 2023-05-30
[108] Unify logging initialization

jpmcfarland:
- use init_logger() and set_log_level() from utils
- move pulling of verbose value to set_log_level()
- update stream_handler log format

Solves #108


wurDevTim (approved):
Changes to the code are small and therefor easy to check.  Also downloaded and ran the code, it looks good. Especially the colored messages are nice.

chStaiger (approved):
Approved!

PR #119, opened 2023-06-06
Updating user docs

chStaiger:
Configurations, paths have changed in the current branch. Update Userdoc pages:

- [x] Python version 3.9
- [x] Command Line Client: uses irods_environment.json and ibridges.json; commmandline parameters changed
- [x] GUI: Add some details to ELN, Amber, Create Ticket and Ticket login

A small bug upon starting the GUI:
```
./iBridges.py
Exception ignored in: <function IrodsConnector.__del__ at 0x12adaa8c0>
  File "/Users/staig001/git-repos/iBridges-Gui/irodsConnector/manager.py", line 59, in __del__
    del self.session
  File "/Users/staig001/git-repos/iBridges-Gui/irodsConnector/manager.py", line 200, in session
    del self._session
AttributeError: _session
```

If we merge this branch into develop, the GitHub pages will be updated and so we should also merge develop into master.


wurDevTim:
It is unchanged in the PR but in the same file: https://github.com/chStaiger/iBridges-Gui/blob/f000132a460d98da6883e4392c7fbda22392e63d/docs/commandlineclient.qmd#L2
Mentions: **files differ**, this is a bit to vague for me. Does it mean file size?

wurDevTim:
Question, in the [getting-started](https://github.com/chStaiger/iBridges-Gui/blob/update-docs/docs/getting-started.qmd) you describe how to configure the iRODS servers to show the free space. In the GUI I see a column `status` next to it, our servers return `none` in this field. Is this some incompatibility with irods 4.3.0 or is there also a rule needed for the status?

wurDevTim:
We know what terms as 'menu bar' mean, but I can image that it is a bit confusing for some users. Is it an idea to add an image?

https://github.com/chStaiger/iBridges-Gui/blob/202311382a0a1e494bdccccf4907d013570a9d27/docs/userdoc.qmd#L56

wurDevTim (approved):
Read the entire files as a double check, it is clear to me what has to be done and I like the references. During my review I fixed a typo and posted two small questions, it already gets my approval.

chStaiger:
> Question, in the [getting-started](https://github.com/chStaiger/iBridges-Gui/blob/update-docs/docs/getting-started.qmd) you describe how to configure the iRODS servers to show the free space. In the GUI I see a column `status` next to it, our servers return `none` in this field. Is this some incompatibility with irods 4.3.0 or is there also a rule needed for the status?

@wurDevTim 
This information comes from the iCAT:
```
cstaiger@rirods4:~$ ilsresc -l irodsRescRepl
resource name: irodsRescRepl
id: 10037
zone: nluu12p
type: random
location: EMPTY_RESC_HOST
vault: EMPTY_RESC_PATH
free space:
free space time: : Never
status: up
info:
comment:
create time: 01485446833: 2017-01-26.16:07:13
modify time: 01542386579: 2018-11-16.16:42:59
context:
parent:
parent context:
```

So yes the sysadmins have to set the information with

```
iadmin modresc status <Value> 
```

README.md
```diff
@@ -31,69 +40,22 @@ Install dependencies with, for example:
...
-    "last_ienv": "irods_environment.json",
-    "ui_tabs": [
-        "tabUpDownload",
-        "tabELNData",
-        "tabDataBundle",
-        "tabCreateTicket"
-    ]
-    "verbose": "info"
-}
+export PYTHON_IRODSCLIENT_DEFAULT_XML=QUASI_XML
```
jpmcfarland, line 48: This variable is set within the iBridges.py script.  Will that work or will the user need to do it this way?

By the way, a Linux user can also add the variable this way:

```bash
...]$ env PYTHON_IRODSCLIENT_DEFAULT_XML=QUASI_XML ./iBridges.py
```

docs/commandlineclient.qmd
```diff
@@ -1,51 +1,47 @@
...
-[iRODS]
-irodsenv = /<path to>/irods_environment.json
-irodscoll =
-irodsresc =
-webdav =
-
-[DOWNLOAD]
-path = /path/to/download/directory
-```
+- Please create a directory named `.irods` in your home directory:
```
jpmcfarland, line 8: Just a suggestion.  This could be generically referred to as `~/.irods`.  In Python, even under Windows, the `~/` is recognized as the user's home directory.  It might even be recognized in a **PowerShell**.

README.md
chStaiger: I tried it without and I run more often into Segmentation Faults ... so just to be sure I would leave it in.

docs/commandlineclient.qmd
chStaiger: Oh that is good to know. I will adjust that!

README.md
alanking: Hi @chStaiger - Where do you see segfaults? In iRODS or somewhere else?

chStaiger: The segfaults are caused by some test .xml files in our iRODS/YODA instance and I can prevent them with the `QUASI_XML` settings. In vanilla iRODS instances I have not encountered the segfaults.

alanking: Okay. If the segfaults are observed in the iRODS server, please file an issue in the irods/irods repository with relevant stacktraces and logs, if possible, so that we can have a look (as it is a serious issue which we need to address). Thanks!

docs/commandlineclient.qmd
jpmcfarland: I just checked under Windows and, indeed, ~/.irods and ~/.ibridges can be used in **PowerShell** to access these locations.  Tab-completion is also possible with this (universal?) path separator.

Also, L11 below could be simplified to a more generic form:

```qmd
        - Windows: c:/Users/\<username\>/.irods/irods_environment.json
```

docs/getting-started.qmd
```diff
@@ -1,11 +1,11 @@
 ---
 title: "iBridges configuration"
 ---
-The iRODS GUI works with any iRODS instance. Here we show how to configure the application to work your specific iRODS instance.
+The iRODS GUI works with any iRODS instance. Here, we show how to configure the application to work your specific iRODS instance.
```
jpmcfarland, line 4: "any iRODS instance" implies to me **any iRODS _version_**.  You may want to clarify that.

docs/getting-started.qmd
```diff
@@ -14,33 +14,38 @@ The iRODS GUI works with any iRODS instance. Here we show how to configure the a
 ```sh
 git clone git@github.com:chStaiger/iBridges-Gui.git
 cd iBridges-Gui
-python3.10 -m pip install -r requirements.txt
+pip install -r requirements.txt
```
jpmcfarland, line 17: I believe `python -m pip install` is the preferred invocation, but functionally either will work.  Also, in addition to `requirements.txt`, those developing/needing to test will need the packages in `requirements_dev.txt`.

docs/getting-started.qmd
```diff
@@ -14,33 +14,38 @@ The iRODS GUI works with any iRODS instance. Here we show how to configure the a
...
+Please see below how to configure the software to connect to your iRODS instance.
 :::
 
 ::: {.callout-note collapse="true"}
 ### Install from zip/release
 ```sh
 unzip iBridges-Gui.zip
 cd iBridges-Gui
-python3.10 -m pip install -r requirements.txt
+install -r requirements.txt
```
jpmcfarland, line 35: Should this be `pip install` and not just `install`?

docs/getting-started.qmd
```diff
@@ -66,40 +71,63 @@ Please see below how to configure the spftware to connect to your iRODS instance
...
-*iBridges* will create its own configuration file in `~/.ibridges/` containing the name of the last *iRODS* environment file used.  This `config.json` file can be updated to control other aspects of *iBridges*.  For example:
+*iBridges* will create its own configuration file in `~/.ibridges/ibridges_config.json` containing the name of the last *iRODS* environment file used.  This `ibridges_config.json` file can be updated to control other aspects of *iBridges*.  For example:
 
 ```json
 {
     "last_ienv": "irods_environment.json",
     "davrods_server": "https://server.fqdn.nl",
+    "eln_token": "<api key>",
+    "amber_token": "<api key>"
     "ui_tabs": [
```
jpmcfarland, line 85: In a recent PR I have added the ability to configure the order of the tabs in the main menu simply by using that order here.  `tabBrowser` and `tabInfo` are added automatically to the beginning and end, respectively, if they are not included here.  A user can put them where they want if they do include them here.

It would be a good idea to add them here explaining at least the existing functionality.  I can then update it in that PR.

docs/getting-started.qmd
```diff
@@ -66,40 +71,63 @@ Please see below how to configure the spftware to connect to your iRODS instance
...
     - `tabUpDownload`: a two-pane upload/download tab
     - `tabELNData`: for the Electronic Lab Notebook, eLabJournal
+    - `tabAmberWorkflow`: up and download data to and from [Amberscript](https://www.amberscript.com/en/) directly from iRODS.
     - `tabDataBundle`: (un)bundle datasets from/to four supported formats
     - `tabCreateTicket`: create iRODS tickets for anonymous access
-- `force_unknown_free_space`: ignore if resources' free space is unannotated
+- `force_transfers`: always write data to iRODS
+- `check_free_space`: checks the metadata *free_space* of resources and only shows resources with a *free_space* larger than 0.
+
+The `check_free_space` and `force_transfers` options are *REQUIRED*. The default values are
```
jpmcfarland, line 109: If the user fails to add them, they will be added accompanied by a warning in the logs.

docs/info.qmd
```diff
@@ -13,18 +13,21 @@ The GUI can be extended by users to implement their own data workflows.
 - `tabBrowser`: (Default) a main table to browse through collections, metadata and ACL manipulation, data deletion and single file up and downloads,
 - `tabUpDownload`: a two-pane upload/download tab
 - `tabELNData`: Uploading data and linking it to an Electronical Labnotebook experiment
-- `tabDataBundle`: (un)bundle datasets from/to four supported formats (depending on extra iRODS microservices)
+- `tabAmberWorkflow`: Up ad downloading data to and from Amberscript from iRODS
+- `tabDataBundle`: (Experimental)(un)bundle datasets from/to four supported formats (depending on extra iRODS microservices)
```
jpmcfarland, line 17: I would not consider this experimental.  Currently, it supports only TAR files with built-in microservices, no compression, no custom microservices.  iRODS will eventually add microservice support for compression and other formats.  When is unknown.

docs/userdoc.qmd
```diff
@@ -1,14 +1,21 @@
...
+iBridges is a GUI to a Data management platform based on iRODS. It exposes iRODS functionality through the graphical interface to the users. We divided the different functions into groups where each function group is represented in an own tab in the GUI. The tabs are configured for the user depending on the user's rights and role in the Data management platform.
+
+The tabs which currently are in use:
+
 - The **Browser** is a tool to browse through iRODS collections and gather information on iRODS data objects
 - The **Up and Download** tab gives tools to up and download large data safely.
 - The **ELN Data upload** tab is made for specific data uploads that belong to an Experiment described in an ElabJournal instance. 
-- In the **Compress/bundle data** tab the user is given tools to bundle (tar) or compress (zip) iRODS data collections and to extract them again.
```
jpmcfarland, line 7: Why remove _bundle_?  _Compress_ functionality is gone, but _bundle_ functionality remains.

docs/userdoc.qmd
```diff
@@ -17,59 +24,113 @@ To browse through your data collections in iRODS click on the **Browser** tab.
    - Click on the `Name`  of a collection or option. This will load the tabs below the table (see below)
```
jpmcfarland, line 24: The whole browser tab and its sub-tabs  have been significantly updated and this documentation should also be.  For instance, the _Resources_ tab is now the _Replicas_ tab and it shows much more information including the resource hierarchy and not just the name of the leaf resource.

utils/context.py
```diff
@@ -229,7 +229,6 @@ def reset(self):
         """Reset existing instances of dynamic class members
 
         """
-        del self.irods_connector
```
jpmcfarland, line 232: This should NOT be removed.  Without it, the conditional at the beginning of the `iBridges.login_function()` will not provide a new connector and there will be remnants of the old session remaining.

utils/context.py
```diff
@@ -238,6 +237,8 @@ def reset(self):
             self.irods_environment.reset()
             filepath = path.LocalPath(self.irods_env_file).expanduser()
             self.irods_environment.filepath = filepath
+        if self.irods_connector:
+            self.irods_environment.reset()
```
jpmcfarland, line 241: This is already done.  I don't understand why this has been added.

utils/context.py
jpmcfarland: I see now that a previous commit modified the `login_function()` functionality incorrectly.  The `irods_connector` MUST be set initially in `main()` and again in `login_function()` _after_ a new session is requested.  If this is not done, the _new_ session will simply be the _old_ one!

irodsConnector/manager.py
```diff
@@ -197,8 +197,9 @@ def session(self) -> session.Session:
 
     @session.deleter
     def session(self):
-        del self._session
-        self._session = None
+        if self._session:
```
jpmcfarland, line 200: This was removed when it was found no longer to be necessary (after removing a circular reference).  It must be determined why it suddenly is needed again.  The reason is important as may potentially cause other issues.

utils/context.py
jpmcfarland: Regardless of the effect removing it has on the code, this method is meant to _reset_ the `Context` instance to as close to its original state as possible.  _Not_ deleting the connector violates that concept.

utils/context.py
jpmcfarland: A recent PR resets the **irods_connector** instead of deleting it.  These two lines were also removed.  I hope that did not restore the bug.

chStaiger: No actually that is very much needed to prevent yet another bug ;) So far all seems really fine.

PR #122, opened 2023-06-08
[121] Adjust Context workflow

jpmcfarland:
- reset irods_connector instead of deleting it
- use debug level logging for deprecation warnings
  - they are only for developers
- remove redundant code snippets
- remove ContextContainer from IrodsModel
  - it was the only one loading upon startup

### DIFF
--- a/iBridges.py
+++ b/iBridges.py
@@ -125,9 +125,6 @@ class IrodsLoginWindow(PyQt6.QtWidgets.QDialog,
         """Check connectivity and log in to iRODS handling common errors.
 
         """
-        # Replacement connector (required for new sessions)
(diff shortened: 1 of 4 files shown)


iBridges.py
```diff
@@ -227,11 +224,10 @@ def main():
     utils.utils.init_logger(THIS_APPLICATION)
```
jpmcfarland, line 224: The idea here is to initialize the logging, instantiate the Context singleton, and immediately set the log level in case it is not the default.  That way all the logging it done as expected.

chStaiger (approved):
Staticmethos in iroxdstreeview are a great idea.

Approved

PR #127, opened 2023-06-12
CLI plugins compatible with Python 3.8

maartenschermer:
Added  `from __future__ import annotations` so `dict` is recognized as a valid type in Python 3.8.


chStaiger (approved):
Nice, short and to the point!!

PR #138, opened 2023-07-03
[137] Remove irods_connector from Context

jpmcfarland:
Invert the relationship between the Context and the IrodsConnector.


chStaiger:
> @chStaiger Shall I remove the deprecated `ContextContainer` as part of this refactor?

Yes, good idea! Thank you.

jpmcfarland:
Commits imminent!

I have not formally tested the `iBrigesCli`, but the GUI works as expected and all the unit tests are passing.  Here is the method to start an interactive session:

```python
import irodsConnector
import utils

irods_conn = irodsConnector.manager.IrodsConnector()
context.irods_env_file = '/Users/jpm/.irods/irods_environment.json'
# irods_conn.password = 'gasfgsafg'
irods_conn.connect()
```

It should be noted that both the `IrodsConnector` and `Context` are singletons now.  Without this, we would be back to passing `ic` all over the place!  Also, certain conventions have been used in all the modules except `iBridgesCli` (`self.conf`, `self.conn`, `self.ienv`, etc.).  For consistency (and simplicity), I would like to bring this to this last module as part of this PR.

Is that possible?

jpmcfarland:
I noticed that `self.irods_conn = None` _after_ `irods_conn = irodsConnector.manager.IrodsConnector()` in the `iBridgesCli` class.  That will need updating.

jpmcfarland:
> I noticed that `self.irods_conn = None` _after_ `irods_conn = irodsConnector.manager.IrodsConnector()` in the `iBridgesCli` class. That will need updating.

I have updated and tested `iBridgesCli.py`, perhaps not rigorously enough, but everything appears to work for single file upload and download.

gui/IrodsExampleTab.py
```diff
@@ -7,58 +9,95 @@
...
     def __init__(self):
+        """
+
+        """
         super().__init__()
         if getattr(sys, 'frozen', False):
             super().setupUi(self)
         else:
             PyQt6.uic.loadUi("gui/ui_files/ExampleTab.ui", self)
+        self.ienv = self.context.irods_environment.config
```
wurDevTim, line 34: Do you agree that is better to refer to the ienv in this files with:  `self.context.irods_environment.config` to avoid local copies which could give problems?

gui/amberWorkflow.py
```diff
@@ -28,12 +31,13 @@ def __init__(self):
             super(amberWorkflow, self).setupUi(self)
         else:
             loadUi("gui/ui_files/tabAmberData.ui", self)
-
+        self.conf = self.context.ibridges_configuration.config
```
wurDevTim, line 34: Local copies

gui/amberWorkflow.py
```diff
@@ -61,11 +65,9 @@ def _initialize_local_model(self, treeView):
     def _initialize_irods_model(self, treeView):
         self.irodsmodel = IrodsModel(treeView)
         treeView.setModel(self.irodsmodel)
-        irodsRootColl = self.context.irods_environment.config.get(
-                'irods_home', '/'+self.context.irods_connector.zone)
-        self.irodsmodel.setHorizontalHeaderLabels([irodsRootColl,
-                                              'Level', 'iRODS ID',
-                                              'parent ID', 'type'])
+        irodsRootColl = self.ienv.get('irods_home', '/'+self.conn.zone)
```
wurDevTim, line 68: Unrelated to this PR, but I notice it due to the line changes, for uniformity either replace with:
`home_coll_str = utils.path.iRODSPath('/', self.conn.zone, 'home')
irods_root_coll = self.ienv.get('irods_home', home_coll_str)`
Or please update the reference in other tabs like example.

gui/dataTransfer.py
```diff
@@ -219,11 +219,10 @@ def __init__(self, localFsPath, coll, upload):
         upload
         """
         super().__init__()
+        self.conf = self.context.ibridges_configuration.config
```
wurDevTim, line 222: local copy

gui/dataTransfer.py
```diff
@@ -307,16 +307,14 @@ def __init__(self, upload, localFS, Coll, totalSize, resource, diff, addFiles, f
         force
         """
         super().__init__()
+        self.conf = self.context.ibridges_configuration.config
```
wurDevTim, line 310: local copy

iBridges.py
```diff
@@ -135,36 +137,34 @@ def login_function(self):
...
             message = 'No network connection to server: %s:%s'
             logging.warning(message, irods_host, irods_port)
             self.envError.setText(message % (irods_host, irods_port))
             self.setCursor(PyQt6.QtGui.QCursor(PyQt6.QtCore.Qt.CursorShape.ArrowCursor))
             return
         self.setCursor(PyQt6.QtGui.QCursor(PyQt6.QtCore.Qt.CursorShape.WaitCursor))
-        self.context.irods_connector.use_icommands = self.use_icommands
-        self.context.ibridges_configuration.config['last_ienv'] = irods_env_file.name
+        self.conn.use_icommands = self.use_icommands
+        self.conf['last_ienv'] = irods_env_file.name
```
wurDevTim, line 161: Here we actively change a local instance of the config, without changing it for the rest of the program. 
I think the single source of truth (singleton) should be updated here.

irodsConnector/manager.py
```diff
@@ -36,11 +34,21 @@ class IrodsConnector:
     _session = None
     _tickets = None
     _users = None
-    _ibridges_configuration = None
-    _irods_env_file = ''
-    _irods_environment = None
     use_icommands = None
 
+    def __new__(cls):
```
wurDevTim, line 39: Do we really have to make the manager a singleton?
In test application I am already instantiating up to 10 managers to speed up data transfers.

gui/IrodsExampleTab.py
jpmcfarland: This is a single, static copy for the life of the class.  I initially chose to use a property as in `irodsConnector.session.Session`, but thought this might be seen as overkill.  The advantages of a property is that it reads the `config` fresh from the `Context` member each time, can be used for runtime documentation, and best of all, _made immutable_.

I also initially made the addition of these particular properties very simple by implementing an abstract base class that could be inherited from.  This, like too many _magic_ properties, proved undesirable as well.  That ABC was the `ContextContainer`.

I can re-implement those constructs if any of the advantages I mentioned are required.

jpmcfarland:
> Request to update the docs, it is the last place where I found a reference to `context.irods_connector`:
> https://github.com/chStaiger/iBridges-Gui/blob/8807392f249a1f5f3e2dcadca1887f85b6ba71ea/docs/developer.qmd#L229C1-L234C82

I am updating.  Should I add anything about the `IrodsConnector` singleton there as well?

gui/amberWorkflow.py
jpmcfarland: Indeed, this is a legacy issue.  I can easily go through these and update them now, or they can wait until I update the style of the particular modules in turn.  Perhaps the latter is best to keep this PR as simple as possible.  A compromise might be to update only those modules that are already being changed and leave the rest for later.

iBridges.py
jpmcfarland: You are very correct!  A property would prevent such an oversight.  I will revert that line.

irodsConnector/manager.py
jpmcfarland: I thought of multiple managers, but a single manager is capable of multiple connections and we can allow _it_ to manage those threads.  If we try to do it, we will inevitably need to add thread safety, something the PRC already does.  I just don't think the current arrangement is ideal for that.

Also, I have been testing alternate clients for our users, and like the PRC, all connections go through the connected catalog server, even if there are separate resource servers.  For those cases, this adds significant load that can be avoided with a single manager.

Eventually, the PRC will add forwarding to the resource server for optimal, multi-threaded transfers.  I think it better to wait for that than to add complexity to iBridges in this way.

In any case, how much improvement are you seeing?  I wonder if it can be repeated with multiple connections from a single manager.  Certainly worth a look?

trel: > Eventually, the PRC will add forwarding to the resource server for optimal, multi-threaded transfers.

PRC 'resource redirect' PR is here... https://github.com/irods/python-irodsclient/pull/458

eyeballs welcome!

test/test_context.py
jtersteegwur, line 1: (windows 11)
All tests in test_context.py pass 

However, the following tests seem to fail, you've not changed this file, but wanted to put it somewhere :)
* test.test_path.test_pure_path_new
* test.test_path.test_pure_path_joinpath
* test.test_path.test_local_path_replace_path

gui/dataTransfer.py
jtersteegwur, line 1: [2023-07-07 15:52:28,812] INFO - UpDownload Diff: ([], [], [...large list of str's...], [])
[2023-07-07 15:52:28,814] INFO - IRODS DOWNLOAD: /RDMacc/home/terst007/frut-->C:/iRods/frut
[2023-07-07 15:52:28,816] INFO - IRODS DOWNLOAD started:
[2023-07-07 15:52:28,816] INFO - IRODS DOWNLOADING /RDMacc/home/terst007/frut/irods-main\scripts\irods\setup_options.py to C:\iRods\frut\frut\irods-main\scripts\irods\setup_options.py
[2023-07-07 15:52:28,852] ERROR - IRODS DOWNLOAD: OBJ_PATH_DOES_NOT_EXIST(None,)
[2023-07-07 15:52:28,852] ERROR - OBJ_PATH_DOES_NOT_EXIST(None,)

gui/IrodsUpDownload.py
jtersteegwur, line 1: windows11:
*uploading a file worked
*downloading a file worked

test/test_context.py
jpmcfarland: I have not run the unit tests for the latest version under Windows or Linux yet.  Linux should be fine, however.

**WILL DO**

gui/dataTransfer.py
jpmcfarland: If this is under Windows, then it is likely to do to the same reason the unit tests fail.  I will have a look.

test/test_context.py
jpmcfarland: Linux is **indeed** fine.

jpmcfarland: Windows is a completely different matter.  The first two tests are related to a `repr` issue, and indeed the `replace_path` error is more complicated and likely causing the problem with downloading collections.

I'll get to it at my earliest convenience.

jpmcfarland: These are fixed with https://github.com/chStaiger/iBridges-Gui/pull/138/commits/87644022ca97f1014d3dceb9878d0c09a2f25bd4.

gui/dataTransfer.py
jpmcfarland: Please try again.  The related tests have been fixed, so it should work now.

iBridges.py
jpmcfarland: As a note, _if_ the `Context.ibridges_configuration` (or the `Context.irods_environment` for that matter) is reinitialized with a new `JSONConfig` at some point, this will break the connection as it will be a new dictionary instance.  This is _not_ done and there is no reason to do it (`JSONConfig`s are `reset()` not replaced between sessions), but a programmer unaware of this _feature_ could cause issues in this way.

test/test_context.py
jtersteegwur: Confirmed! All tests green!

irodsConnector/manager.py
jpmcfarland: > Do we really have to make the manager a singleton?

Also, as discussed offline, when a singleton `IrodsConnector` is used, there is only ever one `irods.Session` object with its pool of connections, that _should_ be usable in a multi-threading environment.

@d-w-moore Is it safe to use in this way?

The alternative to a singleton connector would be _either_ to allow each class that needs its functionality to instantiate a _new_ `IrodsConnector` _or_ to pass the one canonical connector around via arguments as was the original case.

For the multi-connector scenario, there could be up to a dozen or more connectors, each one having to make a _new_ connection to the iRODS server, each with its associated authentication overhead when called upon.  And no telling what kind of thread collisions might happen if we are not careful enough.

As for the single connector scenario, it feels like taking a huge step backwards having to pass it around between classes and methods and functions.  The whole reason we chose to use a singleton Context object with embedded connector in the first place was to avoid exactly this requirement.

jpmcfarland:
I have fixed the `joinpath()` bug under Windows, but this extends to the classes as well:

```python
In [1]: utils.path.iRODSPath("a\\b\\c")
Out[1]: iRODSPath("a\b\c")
```

That should be `iRODSPath("a/b/c")`.  The fix is generic enough for them all.

Coming soon!

gui/dataTransfer.py
jpmcfarland: I have fixed the `joinpath()` bug under Windows, but this extends to the classes as well:

```python
In [1]: utils.path.iRODSPath("a\\b\\c")
Out[1]: iRODSPath("a\b\c")
```

That should be `iRODSPath("a/b/c")`.  The fix is generic enough for them all.

Coming soon!

jpmcfarland: https://github.com/chStaiger/iBridges-Gui/pull/138/commits/c27cbe412ae117c876ffc741cec11f10d6ab9e19 and https://github.com/chStaiger/iBridges-Gui/pull/138/commits/376bf737844814407d1ab9b9072c285b0a551ded should (hopefully!) fix it.

jpmcfarland: https://github.com/chStaiger/iBridges-Gui/pull/138/commits/e775d62c1e2b8788feb9bfd660e4edf982480155 should do it.

Please test again.

test/test_path.py
```diff
@@ -396,7 +396,12 @@ def test_local_path_replace_path(local_path):
         if os.path.isdir(f'{TEMP_DIR}/{dirname}'):
             os.rmdir(f'{TEMP_DIR}/{dirname}')
 
-
+@pytest.mark.xfail(reason="This test is currently broken, but this is the behavior irodsConnector.dataOperations.DataOperation.download_data relies on")
```
jtersteegwur, line 399: The new changes don't work on my platform, I've first added the broken test and will look into solving it

jpmcfarland: It passes on my macOS and Windows machines.  Now, both with _and_ without the decorator.

jtersteegwur:
I've ran all unit tests again today: all tests were green.
as a test, I cloned Quake's source-code and uploaded that using data-transfer. This went without error
Next, I downloaded the uploaded data to a different folder, this seems have went without error as well.

However, if I subsequently download the collection again, new files are detected (this should not be the case).
Somewhere, something broke on windows, but I think this goes beyond the scope of this PR
From my perspective we can merge :)

jpmcfarland:
> However, if I subsequently download the collection again, new files are detected (this should not be the case). Somewhere, something broke on windows, but I think this goes beyond the scope of this PR From my perspective we can merge :)

Definitely make a new issue for this and please share your test details in a PM so I can try to replicate it on Linux and macOS.

PR #144, opened 2023-07-24
Installer script

JTvD:
Following the addition of Nuitka, this PR features:
•	The QT installer framework
•	Background info of how to set everything up for the documentation
•	Background information on signing the installer
Unfortunately I don’t have access to Ubuntu & IOS machines at the moment, I know some minor changes will be required. Therefore I would like to request you to test and if needed debug the flow on those OS’es or keep them out of scope for now. 

**Installer**
There are two issues requesting an installer, this PR makes a good start at solving both (https://github.com/chStaiger/iBridges-Gui/issues/30, https://github.com/chStaiger/iBridges-Gui/issues/34). There are various tools available online. For Windows https://installforge.net/ and https://jrsoftware.org/isinfo.php are recommended, for Linux some others and the same for iOS. An installer per platform is quite a hassle. There are plenty of cross platform options, unfortunately the ones I found are not cheap. During this search I came across the QT installation framework which seems to fit our use case. It’s opensource and supports the same platforms as QT itself. To keep it interesting it’s not used often or well documented. Were used to that by now😉 
Luckily I found this URL which helped me get going: https://scythe-studio.com/en/blog/deploying-app-and-generating-offline-installers-for-windows-qt-installer-framework-tutorial
After installing the installer framework, from the URL below. The general idea is to make the following structure, as done by the installer script now:
-	Installer
    -	Dist
        -	Data, containing the actual code to be shipped
        -	Icon, the icon to show in shortcuts etc.
        -	Meta, containing:
•	The **license file**
•	(optional) **Installerscript.qs** with custom changes to the installer.
•	A **package.xml** with meta information on the application, in the future we might want to retrieve some from git.
-	(optional) **controlscript.js**, to modify the default behavior of the installer (for all apps)
-	**config.xml**, basic configuration for the installer

Note that the install folder attached to the git has a different build up. This is done to avoid duplicate files, the deploy_script puts everything in the right place.

_Other helpful links_
- General description: https://www.qt.io/blog/qt-installer-framework-4.0-released
- Download url: https://download.qt.io/official_releases/qt-installer-framework/ 
- Config installation: https://doc.qt.io/qtinstallerframework/ifw-globalconfig.html 
- More information and some examples: https://doc.qt.io/qtinstallerframework/ifw-overview.html 

I set the paths so the installer copies the program to the 'ApplicationsDir': C:\Program Files on Windows, /opt on Linux and /Applications on OS X. On windows it additionally creates a desktop icon and start menu entry.

**Singing an app**
Just like SSL certificates for websites, there are certificates for application to ensure the user nobody messed with it. You probably already guessed this costs a few euro’s again, in the future we might want to do it. But for now it’s enough to be aware and know that its easy to add. For more information I can recommend: [Cryptography Tools - Win32 apps | Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/seccrypto/cryptography-tools)
The installerframework only has an option to sign macOS applications, as described here:  https://doc.qt.io/qtinstallerframework/ifw-tools.html
On windows the signtool from the windows SDK can be used to sign programs. 
As described here: [How To Sign Qt Apps on Windows for Release (pakkly.com)](https://blog.pakkly.com/how-to-sign-qt-apps-on-windows-for-release/)
During the installation of windows sdk you only have to check the signtool itself, which is 16mb. 
Next these 3 command’s can be run in an elevated (admin) powershell to sign the installer: 

```
$cert = New-SelfSignedCertificate -DnsName https://github.com/chStaiger/iBridges-Gui -CertStoreLocation cert:\LocalMachine\My -Type CodeSigning
$pwd = ConvertTo-SecureString -String "password" -Force -AsPlainText Export-PfxCertificate -cert $cert -FilePath cert.pfx -Password $pwd
& "C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\x64\signtool.exe" sign /fd SHA256 /f cert.pfx /p iBr1dgepw /d "iBridges installer" iBridges_installer.exe
```

For Ubuntu I have not checked the options yet, but I suspect there will be some free solutions.


chStaiger:
@JTvD Can we delete this branch? I think the new Nuitka compilation is oudating this work.

JTvD:
@chStaiger this work is not outdated yet;) I already removed the old Nuitka branch. This one is an example for the next step: the installer. The documentation in this PR is a great starting point for the new installer branch. Therefore I would prefer to keep it for now, at least until we have decided how to move forward.

PR #194, opened 2024-06-04
135 rename

chStaiger:
Add rename/move button to browser


ibridgesgui/browser.py
```diff
@@ -123,6 +123,19 @@ def create_collection(self):
...
         self.load_browser_table()
 
+    def rename_item(self):
+        """Rename/move a collection or data object."""
+        self.error_label.clear()
+        if self.browser_table.currentRow() == -1:
+            self.error_label.setText("Please select a row from the table first!")
+            return
+        item_name = self.browser_table.item(self.browser_table.currentRow(), 1).text()
+        irods_path = IrodsPath(self.session,
```
qubixes, line 133: This seems more complicated than necessary, but I might be missing something. Why is it not simply: `IrodsPath(self.session, self.path_input.text(), item_name)`?

qubixes (approved):
Nice feature! Just one small question on my side.

ibridgesgui/browser.py
chStaiger: Just for the users' convenience. 

The textfield catching the collection path is writeable by the user and hence could not start with '/'.  The whole path then would be prefixed with the session.home and the data would end up somewhere where the user does not expect it. 

So I could make it fail if the input path on top is not well-formatted or I can add the '/'. For now I chose this construction ....

PR #195, opened 2024-06-04
192 search

chStaiger:
Switch search over to TransferDataThread instead of having an own DownloadThread.

### DIFF
--- a/ibridgesgui/gui_utils.py
+++ b/ibridgesgui/gui_utils.py
@@ -6,7 +6,7 @@ from typing import Union
 
 import irods
 import PyQt6
-from ibridges.path import IrodsPath
+from ibridges import IrodsPath, download
 
 from ibridgesgui.config import get_last_ienv_path, is_session_from_config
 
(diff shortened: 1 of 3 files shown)


ibridgesgui/gui_utils.py
```diff
@@ -74,6 +74,26 @@ def prep_session_for_copy(session, error_label) -> pathlib.Path:
     return None
 
 
+def combine_diffs(session, sources: list, destination: Union[IrodsPath, pathlib.Path]) -> dict:
+    """Combine the diffs of several upload or download dry-runs."""
+    diffs = {
```
qubixes, line 79: I would give this a different name. `diff` and `diffs` are quite similar.

ibridgesgui/search.py
```diff
@@ -101,8 +98,9 @@ def load_results(self, results):
         table_data = []  # (Path, Name, Size, Checksum, created, modified)
         for result in results:
             if "DATA_NAME" in result:
-                obj = IrodsPath(self.session,
-                                result["COLL_NAME"] + "/" + result["DATA_NAME"]).dataobject
+                obj = IrodsPath(
```
qubixes, line 101: Instead of the "/" and "+", you can just use a comma?

ibridgesgui/gui_utils.py
```diff
@@ -74,6 +74,26 @@ def prep_session_for_copy(session, error_label) -> pathlib.Path:
     return None
 
 
+def combine_diffs(session, sources: list, destination: Union[IrodsPath, pathlib.Path]) -> dict:
```
qubixes, line 77: This function should probably be made obsolete by implementing it in the ibridges library, but for now a useful feature!

qubixes (changes requested):
I haven't clicked the buttons, but reading to code it looks very reasonable.

ibridgesgui/gui_utils.py
chStaiger: Be my guest ;)

ibridgesgui/search.py
chStaiger: Indeed!

PR #210, opened 2024-06-25
 Add dropdown menu for server templates

chStaiger:
To test you will have to adjust the pyproject toml and allow to use your local ibridges installation.


qubixes (approved):
Tried it out and it works!

At some point we can look at the interface to make the login screen easier for users to understand, but perhaps Maarten also had some feedback already in that direction.

PR #232, opened 2024-07-24
 Saving passwords in config.

chStaiger:
Similar to alias (but without the alias) in the cli.

### DIFF
--- a/ibridgesgui/config.py
+++ b/ibridgesgui/config.py
@@ -145,7 +145,9 @@ def save_current_settings(env_path_name: Path):
 def get_prev_settings():
     """Extract the settings from the configuration."""
     config = _get_config()
-    return config.get("settings", {"settings":{}})
+    if config is None:
+        return {}
+    return config.get("settings", {})
 
 # irods config functions
 
--- a/ibridgesgui/config.py
+++ b/ibridgesgui/config.py
@@ -31,7 +31,7 @@ LOG_LEVEL = {
 
 CONFIG_DIR = Path("~", ".ibridges").expanduser()
 CONFIG_FILE = CONFIG_DIR.joinpath("ibridges_gui.json")
-
+IRODSA = Path("~", ".irods", ".irodsA").expanduser()
 
 def ensure_log_config_location():
     """Ensure the location for logs and config files."""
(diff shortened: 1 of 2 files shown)


ibridgesgui/config.py
```diff
@@ -131,6 +131,21 @@ def _get_config() -> Union[None, dict]:
...
+    config = _get_config()
+    if config is not None:
+        if "settings" not in config:
+            config["settings"] = {}
+        config["settings"][str(env_path_name)] = pw
+        _save_config(config)
+
+def get_prev_settings():
+    """Extract the settings from the configuration."""
+    config = _get_config()
```
qubixes, line 147: You will get an error if `_get_config` returns None, right? Might want to handle that case.

ibridgesgui/config.py
```diff
@@ -131,6 +131,21 @@ def _get_config() -> Union[None, dict]:
...
+    return config.get("settings", {"settings":{}})
```
qubixes, line 148: The second part seems wrong to me, shouldn't it return `{}` instead?

ibridgesgui/login.py
```diff
@@ -86,6 +95,9 @@ def login_function(self):
         try:
             if self.cached_pw is True and self.password_field.text() == "***********":
                 self.logger.debug("Login with %s and cached password.", env_file)
+                with open(IRODSA, "w",  encoding="utf-8") as f:
+                    f.write(self.prev_settings[str(env_file)])
```
qubixes, line 99: I don't know exactly what the env_file object is, but perhaps it would be useful to use the full path? It might not be an issue now, but if something somehow changes the cwd that might be a very difficult issue to track down.

qubixes (changes requested):
Some comments on the code, hope it's helpful!

ibridgesgui/config.py
chStaiger: Done!

ibridgesgui/login.py
chStaiger: The env_file is indeed the full path to the environment.json:

```
env_file = self.irods_config_dir.joinpath(self.envbox.currentText())
```

And in the ibridges-gui.json we store the full path.

PR #282, opened 2024-11-07
Securing the obfuscated password file/folder

joergsteinkamp:
Dear all,
Nice and usefull tool. The obfuscated password needs a better protection in my opinion. I inserted two possibilities, how to create either a folder with permission (700) or a file with (600). Nevertheless, I don't know how this would translate inti the Windows world. See also Stackoverflow where I got the code snippet from:
https://stackoverflow.com/questions/36745577/how-do-you-create-in-python-a-file-with-permissions-other-users-can-write
Kind regards,
Jörg

### DIFF
--- a/ibridgesgui/login.py
+++ b/ibridgesgui/login.py
@@ -23,6 +23,7 @@ from ibridgesgui.config import (
 from ibridgesgui.gui_utils import UI_FILE_DIR
 from ibridgesgui.ui_files.irodsLogin import Ui_irodsLogin
 
+
 def strictwrite(path, flags, mode='0o600'):
     """Create opener for the standard open command to modify the umask."""
     return os.open(path, flags, mode)
--- a/ibridgesgui/login.py
+++ b/ibridgesgui/login.py
@@ -24,7 +24,7 @@ from ibridgesgui.gui_utils import UI_FILE_DIR
 from ibridgesgui.ui_files.irodsLogin import Ui_irodsLogin
 
 
-def strictwrite(path, flags, mode='0o600'):
+def strictwrite(path, flags, mode=0o600):
     """Create opener for the standard open command to modify the umask."""
     return os.open(path, flags, mode)
 
--- a/ibridgesgui/login.py
+++ b/ibridgesgui/login.py
@@ -22,6 +22,8 @@ from ibridgesgui.config import (
 from ibridgesgui.gui_utils import UI_FILE_DIR
 from ibridgesgui.ui_files.irodsLogin import Ui_irodsLogin
 
+def strictwrite(path, flags, mode='0o600'):
+    return os.open(path, flags, mode)
 
 class Login(QDialog, Ui_irodsLogin):
     """Definition and initialization of the iRODS login window."""
@@ -105,7 +107,7 @@ class Login(QDialog, Ui_irodsLogin):
         try:
             if self.cached_pw is True and self.password_field.text() == "***********":
                 self.logger.debug("Login with %s and cached password.", env_file)
-                with open(IRODSA, "w",  encoding="utf-8") as f:
+                with open(IRODSA, "w",  encoding="utf-8", opener=strictwrite) as f:
                     f.write(self.prev_settings[str(env_file)])
 
                 session = Session(irods_env=env_file)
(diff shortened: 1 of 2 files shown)


chStaiger:
Thank you for the code contribution, it is. indeed very useful. We will test whether it will also work in our windows environments. And we will also adjust the iBridges CLI in our API package accordingly.

chStaiger:
@joergsteinkamp  I started our linting and it finds some python errors. E.g. the `os` module is not imported etc. Could you have a look at this please?

chStaiger:
Thanks for the update. Unfortunately the linters are still complaining.
You can check it yourself locally with 

```
ruff ibridgesgui
pylint ibridgesgui
```

joergsteinkamp:
Sorry, I missed inserting the 'import os' in login.py in my github repo. It's now in there. Also the docstring in imperative: 'Creates' and 'Creating' was not accepted.

ibridgesgui/login.py
```diff
@@ -22,6 +23,9 @@
 from ibridgesgui.gui_utils import UI_FILE_DIR
 from ibridgesgui.ui_files.irodsLogin import Ui_irodsLogin
 
+def strictwrite(path, flags, mode='0o600'):
```
chStaiger, line 26:
```suggestion

def strictwrite(path, flags, mode='0o600'):
```

qubixes (approved):

chStaiger:
Thank you for the update. I tested your new feature on Mac, but I run into an error:

```
ERROR:ibridges-gui:Failed to login: TypeError("'str' object cannot be interpreted as an integer")
  File "/Users/user/test-git/iBridges-GUI/ibridgesgui/login.py", line 113, in login_function
    with open(IRODSA, "w",  encoding="utf-8", opener=strictwrite) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/test-git/iBridges-GUI/ibridgesgui/login.py", line 29, in strictwrite
    return os.open(path, flags, mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'str' object cannot be interpreted as an integer
```

chStaiger:
Thanks for the update. I tried the new feature on Mac but I run into this error:

```
ERROR:ibridges-gui:Failed to login: TypeError("'str' object cannot be interpreted as an integer")
  File "/Users/user/test-git/iBridges-GUI/ibridgesgui/login.py", line 113, in login_function
    with open(IRODSA, "w",  encoding="utf-8", opener=strictwrite) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/test-git/iBridges-GUI/ibridgesgui/login.py", line 29, in strictwrite
    return os.open(path, flags, mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'str' object cannot be interpreted as an integer
```

It might also be nice to have a unit test for this function. Would you like to write that test or would you like us to do it?

ibridgesgui/login.py
```diff
@@ -23,6 +24,10 @@
 from ibridgesgui.ui_files.irodsLogin import Ui_irodsLogin
 
 
+def strictwrite(path, flags, mode='0o600'):
```
chStaiger, line 27:
```suggestion
def strictwrite(path, flags, mode=600):
```

chStaiger (changes requested):
Something seems to be wrong with the mode flag in `os.open`:

```
In [4]: ?os.open
Signature: os.open(path, flags, mode=511, *, dir_fd=None)
Docstring:
Open a file for low level IO.  Returns a file descriptor (integer).
```

joergsteinkamp:
Ok, on Linux I had trouble with integers looking like permissions, the mode was more readable for me in the octal version. 511 is 0o777 in octal, which is the default and would give the default as access rights. 0o600 would be 384 and 0o700 448, respectiively. Use print(0o700) to convert it to integers and oct(511) vice versa.

ibridgesgui/login.py
chStaiger, line 27:
```suggestion
def strictwrite(path, flags, mode=0o600):
```

chStaiger (approved):
Thanks for this very useful and nice contribution!

PR #286, opened 2024-11-24
swap md set to update, fix md delete bug

chStaiger:
The browser has now a real Metadata update button which exchanges the selected metadata with the values provided in the text fields.

Unfortunately I could not. make use of our new coll backend functionality, since there you can only change one item of the metadata triple at a time. The GUI asks all information at on time point. But adding those changes subsequently can lead to hitting already existing metadata. So I implemented an own function which should do the trick.

All feedback is welcome.

### DIFF
--- a/ibridgesgui/browser.py
+++ b/ibridgesgui/browser.py
@@ -2,6 +2,7 @@
 
 import logging
 import sys
+from typing import Optional, Union
 
 import irods.exception
 import PyQt6.QtCore
@@ -9,7 +10,6 @@ import PyQt6.QtGui
 import PyQt6.QtWidgets
 import PyQt6.uic
 from ibridges import IrodsPath
-from ibridges.meta import MetaData
 from ibridges.permissions import Permissions
 from ibridges.util import obj_replicas
 
@@ -26,7 +26,7 @@ from ibridgesgui.ui_files.tabBrowser import Ui_tabBrowser
 class Browser(PyQt6.QtWidgets.QWidget, Ui_tabBrowser):
     """Browser view for iRODS session."""
 
-    def __init__(self, session, app_name):
+    def __init__(self, session, app_name: str):
         """Initialize an iRODS browser view."""
         super().__init__()
         if getattr(sys, "frozen", False):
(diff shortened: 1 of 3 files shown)


qubixes:
@chStaiger I understand the problem: (k1, v, u) -> (k, v, u) and (k, v, u) -> * (k1, v, u) are two operations that are individually incompatible, but they could be performed at the same time. However, I don't understand how the code in this PR solves this? You would need to create a "diff" between the old and new metadata, add all the new items, then remove all the old items (and roll-back if something happens halfway).

chStaiger:
You are right, if we can restrict the users to only change one part of the metadata triple at a time, that would make it much easier and would be a true "edit".

In the current implementation I check whether the new metadata triple is well-formed, i.e. the key is new or the old one, the value is present (either new or old) and the new triple does not exist. Then I add it as new metadata and remove the old metadata triple. So in that sense it is a "replace".

Which one would we like to implement? The "edit" with the restriction of changing only one part of the metadata triple; or the "replace"?

@JTvD Do you have an idea which one would suit your use cases better?

qubixes:
Oh, I see. That was not the problem I was thinking about. Updating both the key and value at the same time would be a problem, yes. We could fix that on the API side though. We just need a set method for the MetaDataItem? That would not be so complicated.

chStaiger:
> Oh, I see. That was not the problem I was thinking about. Updating both the key and value at the same time would be a problem, yes. We could fix that on the API side though. We just need a set method for the MetaDataItem? That would not be so complicated.

Ah, I see. You thought I wanted to change two metadata triples at a time. That is indeed not the case.

If we can do that for the full triple, that would be great. That would solve the problem here. But you would then have to solve the problem in the API that an intermediate step to change the might result in an existing metadata entry, while the end result is new.

qubixes:
Why would there be an intermediate value?

chStaiger:
Assume you have these two metadata entries:

```
key1, val1, units1
key2, val1, units1
```

Now you want to change the first one to `key2, val1, units3`. It makes a difference in which order you replace the key, value or units. If you start with the key, you will create an identical metadata entry to the second one.

qubixes:
But you can just add the new triple and then remove the old, right? There is not need to do them separately?

chStaiger:
> But you can just add the new triple and then remove the old, right? There is not need to do them separately?

Yes, that is exactly what I did on the code and why I could not use our current implementation of the metadata edits in the API.

PR #290, opened 2024-12-05
bugfix class names in ui files

JTvD:
Heey,

Somewhere in the editing process the class names of the Ui files have been removed.
To my surprise they are correct in the '.py' files, no clue what went wrong there but it's an easy fix.

Hope we can get this merged soon (I am compiling executables and this breaks everything)


chStaiger:
Thanks for spotting that!

Issue #293, opened 2024-12-23
refactor pyqt to pyside


JTvD:
PyQt6 has a GPL or commercial license
PySide6 comes with a LGPL license

Even for the opensource project it is preferred to switch to the less restrictive license.
Among other things it gives users the opportunity to add connections to other services without having to share that code as well.
However, where possible I would still like to recommend to make as much as possible opensource!
Will all benefit from the collaboration.

PR #294, opened 2024-12-23
refactored pyqt6 to Pyside6

JTvD:
Implements #293. There have been a few attempts to move to pyside before, unfortunately there are a few tricky changes between pyqt and pyside. 
One of them is the custom UiLoader needed to load UI files in pyside6 like done in pyqt. I have not found any function in the pyside6 library that could be used as alternative.
A second challenge is that the QPlainTextEditLogger, which inherits the 'PySide6.QtCore.QObject' and must therefore be running on the main thread was emitting signals to the qtextedit. This is not needed as a class that has a reference to a widget must be on the mainthread. 

Regardless,  I changed it is because pyside6 signals do not directly slot into the QT slot like the signals of pyqt6. 
there are quite a few topics about this on [stackoverflow ](https://stackoverflow.com/questions/75617527/why-do-similar-signal-slot-connections-behave-differently-after-pyqt5-to-pyside6). For as far as I saw this is the only place where it was done, if it is needed else where a simple pyside6 slot that passes the string to the widget suffices. 

Based on a few hours of testing it seems to be working. Nevertheless it is recommended to test the code thoroughly.

Note, the code to create the new ui_files follows in a second pull request.


JTvD:
#294

qubixes:
@JTvD Thanks a lot! I can confirm that everything seems to work that I have tried so far (which is not much).

chStaiger (approved):
Great Work! I just tested the main look and feel and the buttons. All works and finally pyside also instlls nicely on my computer. 

Thank you very much!

PR #296, opened 2024-12-24
Nuitka stand alone executables

JTvD:
It is unfortunate that when the ‘--onefile’ flag is passed to a python package the executable is first unpacked into a temporary directory before it is started. Therefore I recommend to build the executable as a folder. One big advantage of Nuitka over pyinstaller is that it compiles the python code to C where possible. In my tests on Windows this executable is 5x faster than the one build by pyinstaller. The price to pay is some compilation time, sounds like a good trade to me.

There is some weird behavior, when building the executable the first time I got the error '"irods_authentication_scheme": "irods.auth.pam_password" does not exist'.
To solve this I forced nuitka to include the entire python irods package (minus the test folder) in the build:
"--include-package=irods --nofollow-import-to=irods.test"
With this change all functionalities work. It is not clear to me why this specific import is not followed by default.

Additional changes: 
- At the end of the build script the naming is updated to translate '__main__/__main.exe' into 'ibridgesgui/ibridgesgui.exe'
- Added a script to convert the '.ui' files into '.py' files
- Updated the references to the 'LOGO_DIR' in both the gui_utils.py as the '.py' ui files. This simplifies the folder tree
  - From: folder with exe > ibridgesgui/icons/*.png
  - To: folder with exe > icons/*.png

The code has been tested on IOS and ubuntu but its always recommended to give it a check. The majority of my testing was done on windows in an effort to make iBirdges available on all WUR machines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ __pycache__/
 build/
 develop-eggs/
 dist/
-output/
+ibridgesgui_dist/
 downloads/
 eggs/
 .eggs/
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ __pycache__/
 build/
 develop-eggs/
 dist/
+output/
 downloads/
 eggs/
 .eggs/
--- a/README.md
+++ b/README.md
@@ -68,10 +68,27 @@ The git repository contains a generic *iRODS* graphical user interface.  The iRO
  ## Building Executables
 
 We offer build scripts in case you would like to compile iBridges and distribute the executable.
-Simply run:
 
 ```
-python build_tools/build_script.py
+python3 build_tools/build_script.py -h
+
+usage: build_script.py [-h] [--debug_exe] [--rem_venv]
+                       [--code_folder CODE_FOLDER] [--ui_folder UI_FOLDER]
+                       [--icons_folder ICONS_FOLDER] [--venv VENV]
+
+iBridges-Gui exe creator.
+
+options:
+  -h, --help            show this help message and exit
+  --debug_exe           Build executable with debug console
+  --rem_venv            remove virtual environment
+  --code_folder CODE_FOLDER
+                        Full path to the directory with code
+  --ui_folder UI_FOLDER
+                        Full path to the directory with ui files
+  --icons_folder ICONS_FOLDER
+                        Full path to the directory with the icons
+  --venv VENV           Full path to virtual python environment
 ```
 
 This will create a virtual environment `venv` in your iBridges folder. Next to that you will find a new folder `build` and `output`. The executable can be found and started like this for Mac and Linux:
--- a/README.md
+++ b/README.md
@@ -64,7 +64,25 @@ The git repository contains a generic *iRODS* graphical user interface.  The iRO
   ```bash
   python ibridgesgui/__main__.py
   ```
- 
+
+ ## Building Executables
+
+We offer build scripts in case you would like to compile iBridges and distribute the executable.
+Simply run:
+
+```
+python build_tools/build_script.py
+```
+
+This will create a virtual environment `venv` in your iBridges folder. Next to that you will find a new folder `build` and `output`. The executable can be found and started like this for Mac and Linux:
+
+```
+./output/ibridgesgui/ibridges_gui.bin
+```
+
+For Windows simply click on the file `output/ibridgesgui/ibridges_gui.exe`.
+
+Please note, that this feature is still in experimental state.
  
 
 ## Authors
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ options:
   --venv VENV           Full path to virtual python environment
 ```
 
-This will create a virtual environment `venv` in your iBridges folder. Next to that you will find a new folder `build` and `output`. The executable can be found and started like this for Mac and Linux:
+This will create a virtual environment `venv` in your iBridges folder. Next to that you will find a new folder `build` and `ibridgesgui_dist`. The executable can be found and started like this for Mac and Linux:
 
 ```
 ./output/ibridgesgui/ibridges_gui.bin
--- a/build_tools/build_script.py
+++ b/build_tools/build_script.py
@@ -67,16 +67,12 @@ def create_exe():
                                                        icons_folder.name))
 
     # Step 5, rename the distribution folder and file
-    shipping_folder = Path('dist').joinpath('ibridges-gui')
+    shipping_folder = Path('output/ibridgesgui')
     if Path(shipping_folder).exists():
         rmtree(shipping_folder, ignore_errors=True)
     shipping_folder.parent.mkdir(parents=True, exist_ok=True)
     Path('__main__.dist').rename(shipping_folder)
-    # Windows
-    if system()[0].upper() == "W":
-        Path(shipping_folder).joinpath('__main__.exe').rename((shipping_folder).joinpath('ibridges-gui.exe'))
-    else:  # Ubuntu/IOS
-        Path('dist').joinpath('__main__').rename((shipping_folder).joinpath('ibridges-gui'))
+    Path(f'{shipping_folder}/__main__.exe').rename(f'{shipping_folder}/ibridges.exe')
 
 
 if __name__ == "__main__":
--- a/build_tools/build_script.py
+++ b/build_tools/build_script.py
@@ -41,7 +41,8 @@ def create_exe():
             venv_activate = f"source {venv_activate}"
         uipy.run_cmd(f"{python} -m venv {venv}")
         uipy.run_cmd(f"{venv_activate} {cmd_sep} python -m pip install --upgrade pip")
-        uipy.run_cmd(f"{venv_activate} {cmd_sep} pip install -r requirements.txt")
+        uipy.run_cmd(f"{venv_activate} {cmd_sep} pip install .")
+        uipy.run_cmd(f"{venv_activate} {cmd_sep} pip install .[deploy]")
     else:
         venv_activate = f"\"{str(venv_activate)}\""
 
--- a/build_tools/build_script.py
+++ b/build_tools/build_script.py
@@ -1,8 +1,8 @@
 """Build executable with Nuitka."""
+import argparse
 from pathlib import Path
 from platform import system
 from shutil import copytree, rmtree
-import argparse
 
 import ui_to_py as uipy
 
@@ -14,8 +14,9 @@ def replace_folder(source: Path | str, destination: Path | str):
 
 def create_exe(args):
     """Create the exe file for the application.
-        args : argparse.Namespace
-            arguments passed to the function
+
+    args : argparse.Namespace
+        arguments passed to the function
     """
     (python, cmd_sep) = uipy.os_specific_settings()
 
@@ -77,15 +78,20 @@ def create_exe(args):
 if __name__ == "__main__":
     default_code_folder = Path.cwd().joinpath('ibridgesgui')
     parser = argparse.ArgumentParser(description="iBridges-Gui exe creator.")
-    parser.add_argument('--debug_exe', action="store_true", help='Build executable with debug console')
-    parser.add_argument('--rem_venv', action="store_true", help='remove virtual environment')
+    parser.add_argument('--debug_exe', action="store_true",
+                        help='Build executable with debug console')
+    parser.add_argument('--rem_venv', action="store_true",
+                        help='remove virtual environment')
     parser.add_argument('--code_folder', default=default_code_folder, type=Path,
                         help='Full path to the directory with code')
-    parser.add_argument('--ui_folder', default=default_code_folder.joinpath('ui_files'), type=Path, required=False,
+    parser.add_argument('--ui_folder', default=default_code_folder.joinpath('ui_files'),
+                        type=Path, required=False,
                         help='Full path to the directory with ui files')
-    parser.add_argument('--icons_folder', default=default_code_folder.joinpath('icons'), type=Path, required=False,
+    parser.add_argument('--icons_folder', default=default_code_folder.joinpath('icons'),
+                        type=Path, required=False,
                         help='Full path to the directory with the icons')
-    parser.add_argument('--venv', default=Path.cwd().joinpath('venv'), type=Path, required=False,
+    parser.add_argument('--venv', default=Path.cwd().joinpath('venv'),
+                        type=Path, required=False,
                         help='Full path to virtual python environment')
     args = parser.parse_args()
     create_exe(args)
--- a/build_tools/build_script.py
+++ b/build_tools/build_script.py
@@ -65,10 +65,9 @@ def create_exe(args):
                                                             args.icons_folder.name))
 
     # Step 5, rename the distribution folder and file
-    shipping_folder = Path('output/ibridgesgui')
+    shipping_folder = Path('ibridgesgui_dist')
     if Path(shipping_folder).exists():
         rmtree(shipping_folder, ignore_errors=True)
-    shipping_folder.parent.mkdir(parents=True, exist_ok=True)
     Path('__main__.dist').rename(shipping_folder)
     for file in Path(shipping_folder).glob('__main__.*'):
         suffix = file.suffix
--- a/build_tools/build_script.py
+++ b/build_tools/build_script.py
@@ -29,7 +29,7 @@ def create_exe(args):
     # windows
     if "windows" in system().lower():
         venv_activate = args.venv.joinpath('Scripts', 'activate.bat')
-    else:  # Ubuntu/IOS
+    else:  # Ubuntu/Mac OS
         venv_activate = args.venv.joinpath('bin', 'activate')
     if (not venv_activate.exists()) or (not venv_activate.is_file()):
         venv_activate = f"\"{str(venv_activate)}\""
--- a/build_tools/build_script.py
+++ b/build_tools/build_script.py
@@ -2,18 +2,9 @@
 from pathlib import Path
 from platform import system
 from shutil import copytree, rmtree
+import argparse
 
 import ui_to_py as uipy
-
-# Build config and file paths
-code_dir = Path("ibridgesgui")
-icons_folder = Path.cwd().joinpath(code_dir, 'icons')
-ui_folder = Path.cwd().joinpath(code_dir, 'ui_files')
-venv = Path.cwd().joinpath('venv')
-# Debug mode shows the console (command prompt with the logging)
-debug_mode = True
-remove_venv = False
-# ---------------------------------------------------------#
 
 
 def replace_folder(source: Path | str, destination: Path | str):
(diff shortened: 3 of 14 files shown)


build_tools/build_script.py
```diff
@@ -0,0 +1,79 @@
...
+                   icons_folder.parent.parent.joinpath('__main__.dist',
+                                                       icons_folder.name))
+
+    # Step 5, rename the distribution folder and file
+    shipping_folder = Path('output/ibridgesgui')
+    if Path(shipping_folder).exists():
+        rmtree(shipping_folder, ignore_errors=True)
+    shipping_folder.parent.mkdir(parents=True, exist_ok=True)
+    Path('__main__.dist').rename(shipping_folder)
+    Path(f'{shipping_folder}/__main__.exe').rename(f'{shipping_folder}/ibridges.exe')
```
qubixes, line 75: Shouldn't this be ibridges-gui.exe? Also, for non-windows, there probably shouldn't be the `.exe` suffix, right?

qubixes (commented):
It's very nice to be able to create a (fast) executable, thanks for implementing it! I currently don't have a Python/windows setup, but I can if needed test it out later. 

It would be nice if the executable is built automatically with every release. That should be possible with github Actions, right?

Btw, the setup with the venv seems rather complicated, but I suppose that's how Nuitka is supposed to work? I was trying to figure it out from their documentation, but it seems a little sparse.

build_tools/build_script.py
JTvD: Good catch, this is why it would be nice to test this on a different OS. Unfortunately I only have headless Ubuntu machines. Also renamed the exe to 'ibridges-gui.exe'.

qubixes:
Yes, it would be nice to build the executables every release automatically with GH actions, but I agree with you that that can wait. It looks like we're already quite close though (thanks to this PR)! I would say mac / ubuntu is a lower priority, since most users will only have to use `pip install`, but since the infrastructure is there, we might as well build them.

JTvD:
@qubixes is also my preference to build the exe for all three operating systems. I already did some homework, the git actions are described in the documentation of Nuitka:  [github actions](https://nuitka.net/user-documentation/use-cases.html#building-with-github-workflows) That is an easy setup, personally prefer to have the script to build it locally (this PR) as well so we can debug the builds and test new functionality.

chStaiger:
I tested the build script on Mac and Ubuntu20.04:
- Mac
   - Mac bulds a `__main__.bin` instead of an `exe`. This is fixed in  the commit [560d703](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/560d703ed12209d565d477e470f8dfb3b057a12f)
   - Mac can only execute the `bin` file on the command line, i.e. right click -> execute with terminal. To create a double-clickable icon on the desktop or in the folder one has to provide a `command` file which then can also have a logo.
   - The logo on the welcome tab in the iBridgesGUI is not shown
- Ubuntu20.04
  - I could not run the script since the defult python compiler on ubuntu is 3.8. I also have python 3.10 installed which should work. However the build script needs to be adjusted so that python3.10 or another version can be passed as a parameter to the build process

I could not test the build on Windows due to python not being able to be installed on research cloud. I need a different environment.

build_tools/build_script.py
```diff
@@ -0,0 +1,83 @@
+"""Build executable with Nuitka."""
+from pathlib import Path
+from platform import system
+from shutil import copytree, rmtree
+import platform
```
JTvD, line 5: @chStaiger see line 3, please use only one of the two

chStaiger: Aye aye ;)

chStaiger:
Some comments on the coding:

- `system()[0].upper() == "W"`: I think it is better to use `"windows" in system().lower()`. It is more robust, just in case ...
- The function `run_cmd` is now throwing a very generic exception which is unnecessary. The function can directly do a sys.exit() since we are not going to capture the exception and do then something else when it fails. It would also be nice if the function could return the STDOUT if successful to make it more versatile in case we want to create a local installer script.
- Can we make the global variables in `build_script.py` configurable? We can keep the current settings as defaults.
- The deug mode is nice, but certainly needs to be made configurable ;)
- Rename the directory `output` to `ibridges_bin` or something like that

Some unexpected features:

- The venv cannot be reused  at least on my mac:
  `commandline error: /bin/bash: /Users/user/git-repos/iBridges-Gui/venv/bin/activate: Permission denied`
   When rerunning the script can you remove all files and folders that were created in a previous run beforehand?

- To also allow for local installs the default install directory for Mac should be `/Users/user/Applications`, for linux I am not so sure ... @qubixes do you know where applications are installed by default on linux?

- The script currently does not work for linux since the default command `python` or `python3` refers to an older version of python than iBridges is built for.

qubixes:
I'm trying to build it on Linux, and I get `FATAL: pyside6: Error, there is no Qt plugin family 'styles'.` Any ideas?

JTvD:
Thanks for testing on the different operating systems. I am still processing the elaborate feedback and testing on a few systems myself, doing my best to process everything this week.  What I can already mention is that I noticed some misalignment on the scope of this PR in the messages. The idea of this script is to create a way for developers to test if their code works in the Executable. Moving files to a different folder like `/Users/user/Applications` or creating scripts/shortcuts is outside of the scope. It requires OS specific integrations, correctly setting the work directory (Qt uses relative paths) and even some certificates to stop virus scanners from deleting the program.  An old attempt at this can be found in [PR installer](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/144).

JTvD:
@qubixes today I have been testing on a few ubuntu distributions. On some I could reproduce your issue. Unfortunately I could not find much information on the plugins and if we still need them with pyside6. 

As a test I removed them and the executable looks and functions the same on my windows and ubuntu machines. Can you give removing `--include-qt-plugins=sensible,styles \` a try? If it also looks the same for you I think we can safely remove the plugins.

JTvD:
It’s been a busy week, especially testing on all the different machines was interesting. It works on all the configurations I have available, and I believe all the feedback has been processed. Therefore, I would like to ask you to give it another look:

Changelog: 
- @chStaiger updated the renaming of the files on IOS in [account for different underlying os (mac)](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/560d703ed12209d565d477e470f8dfb3b057a12f). This did not yet covert Ubuntu. commit [suffix independent exe renaming](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/319b0acaf490e6728b4f12789433a45135016172) solves the executable renaming on all operating systems
- [robust os checks](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/561e230e94cee89267e77b9d300b66a156208412) and  [exit at error and bugfix](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/083053d80321ef2ebbd2276dcecd203e09959f77) address the checks off the operating system and stopping the program in case a command failed.
- [build for the python version calling the script](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/8cf18d394f4389f35fa95965611b2907463faf94) ensures python builds the executable for the python version the script is started with. This is more robust than the proposed solution of adding an argument, at least on my machine.
- [option to remove venv](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/ed02852fd88a4ae00a963ea41dce67d83c5c25d7) Adds the option to remove the virtual environment. Solves the problem that not all operating systems allow other processes to reuse it
- [removed plugins](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/ed46774d98704520974021cf86378952e4be0dfe) removed the QT plugins, it looks like this works robustly.
- [return subprocess output](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/58ae96f484aa6ad9f1b847b385030bed07896615) returns the output of the subprocess for future work.
- [converted all globals into arguments](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/ab3892fb5617e3bca6959c353e50fbbfd72ebd4d) makes the script configurable, left the parser in the `if __name__ == "__main__":` to leave this script callable from another script. I expect we will either move the parser to a different script or extend this script with an installer.

qubixes:
@JTvD Thanks for adapting the code. Functionally, I have two comments:

- Is it possible to put the executable in the build directory? Currently there are both the `output` and `build` directory and I would normally expect the executable to be in the `build` one. 
- When executing the binary from a different directory, it doesn't load the images. So `output/ibridgesgui/ibridges_gui.bin` -> no images, and `cd output/ibridgesgui; ./ibridges_gui.bin` does not.

JTvD:
@qubixes based on your reply it reads as if the executable builds now and works. That's good progress.
To answer your questions: 
- By default nuitka makes two folder: a **build** folder containing temporary files, which it reuses the next time you build the executable. shortening the build time and a **dist** folder containing the executable. At the end of the build script the **dist** for and **__main__.exe** are renamed to ibridgesgui, since the code directory is also called ibridgesgui, I added the **output** folder to prevent conflicts.

- Both PYQT and PYSIDE use relative folder paths, on itself that is a nice feature because it enables you to copy the executable to a different location and still be able to start it. However, it does require you to set the working directory to the right folder when starting the program as you also noticed.  Next to this we want shortcuts, to see the application in the 'programs' list etc. addressing these issues will double, possible even triple this PR in size.  Therefore I proposed to solve it a following PR, focusing on the installation.

qubixes:
@JTvD Okay, I don't mind it too much for the first point. It's only developers that need to deal with that anyway. Perhaps we should start some developer documentation for these things?

For the second point, I understand that relative paths are convenient. However, it would be nicer if the path was relative to the location of the executable and not relative to the current working directory. Are you sure this is not an easy fix somewhere? If it is as you say a lot of work, it can be dealt with another time.

build_tools/build_script.py
```diff
@@ -0,0 +1,91 @@
...
+    # Step 1a,(optional) remove the venv to prevent access issues.
+    if args.rem_venv:
+        rmtree(args.venv, ignore_errors=True)
+    args.venv.mkdir(exist_ok=True)
+
+    # Step 1b, Create the venv if needed
+    # windows
+    if "windows" in system().lower():
+        venv_activate = args.venv.joinpath('Scripts', 'activate.bat')
+    else:  # Ubuntu/IOS
```
qubixes, line 32:
```suggestion
    else:  # Ubuntu/Mac OS
```

build_tools/ui_to_py.py
```diff
@@ -0,0 +1,117 @@
...
+    else:  # Linux, ensure it uses python 3
+        cmd_sep = ";"
+        python = f"python{python_version}"
+    return (python, cmd_sep)
+
+
+def ui_to_py(ui_folder: Path, venv_activate: str, cmd_sep: str):
+    """Convert the .ui files to .py files.
+
+    pyside6-uic gui/MainWindow.ui -o gui/MainWindow.py
```
qubixes, line 54: This an example of what you would normally run? Perhaps this could be clearer.

pyproject.toml
```diff
@@ -44,8 +43,14 @@ test = [
...
 
+deploy = [
+    "Nuitka",
+    "imageio",
+]
+
 all = [
-    "ibridgesgui[test]"
+    "ibridgesgui[test]",
+    "ibridgesgui[deploy]",
```
qubixes, line 53:
```suggestion
    "ibridgesgui[test,deploy]",
```

requirements.txt
qubixes, line 1: Perhaps this PR hasn't been rebased yet?

qubixes (changes requested):
I only have some small comments. Given that this is only for creating the executable, it doesn't come as precise. I would personally prefer numpy style docstrings, but it's not so important. After the comments in my review are addressed, I will approve the PR. Thanks for the hard work!

build_tools/build_script.py
JTvD: There are two more places where this should be updated for consistency, added all in commit [rename IOS to Mac OS](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/a5a55130ad69a2aaa9e31fc1bc286cd724acbc8b)

build_tools/ui_to_py.py
JTvD: Yess, addressed in: [example command](https://github.com/iBridges-for-iRODS/iBridges-GUI/pull/296/commits/9769c1d1ec3adc79902248c348a66f5b284defbb)

requirements.txt
JTvD: Indeed, the rebase did not remove the file so I did it in a new commit. Also suprised me that I had to force the push after completing the rebase "Successfully rebased and updated refs/heads/nuitka."

JTvD:
> @JTvD Okay, I don't mind it too much for the first point. It's only developers that need to deal with that anyway. Perhaps we should start some developer documentation for these things?
> 
> For the second point, I understand that relative paths are convenient. However, it would be nicer if the path was relative to the location of the executable and not relative to the current working directory. Are you sure this is not an easy fix somewhere? If it is as you say a lot of work, it can be dealt with another time.

There is a easy way, I am not thrilled about it but we can set the working directory before we load the GUI.
Can you check if adding:
```    
 # Set the working directory to the directory of the current file
os.chdir(os.path.dirname(os.path.abspath(__file__)))
``` 
before:  https://github.com/iBridges-for-iRODS/iBridges-GUI/blob/77c0a9036aa944bb26d33339c8355e253d525f98/ibridgesgui/__main__.py#L188
Works on your system?

qubixes:
> > @JTvD Okay, I don't mind it too much for the first point. It's only developers that need to deal with that anyway. Perhaps we should start some developer documentation for these things?
> > For the second point, I understand that relative paths are convenient. However, it would be nicer if the path was relative to the location of the executable and not relative to the current working directory. Are you sure this is not an easy fix somewhere? If it is as you say a lot of work, it can be dealt with another time.
> 
> There is a easy way, I am not thrilled about it but we can set the working directory before we load the GUI. Can you check if adding:
> 
> ```
>  # Set the working directory to the directory of the current file
> os.chdir(os.path.dirname(os.path.abspath(__file__)))
> ```
> 
> before:
> 
> https://github.com/iBridges-for-iRODS/iBridges-GUI/blob/77c0a9036aa944bb26d33339c8355e253d525f98/ibridgesgui/__main__.py#L188
> 
> Works on your system?

Yes, that works on my system. I suppose it's because the LOGO_DIR and UI_FILE_DIR are set in gui_utils.py.

qubixes:
> Partly, but these are not the problem. The challenge lies inthe `.ui` the ui_files, they use relative references to the working directory:
> 
> https://github.com/iBridges-for-iRODS/iBridges-GUI/blob/77c0a9036aa944bb26d33339c8355e253d525f98/ibridgesgui/ui_files/tabBrowser.py#L112
> 
> I propose to add:
> 
> ```
>  # Set the working directory to the directory of the current file
> os.chdir(os.path.dirname(os.path.abspath(__file__)))
> ```
> 
> in the main.py to resolve this request, @qubixes is that okay?
> 
> Also thought about the documentation but as there are more changes coming and I don't see a 'for developer' section yet I don't think it's wise to introduce it now in this branch...

Yes, from my side it sounds like a good solution.

For the developer documentation, we could also use the Wiki, then it can be independently done from this PR (after it is merged of course).

JTvD:
Done, this should be everything for the executable. 
Learned a lot again, working with so many different operating systems and python versions remains tricky.

Think it's also good to link to documentation from QT and other packages about which versions of which operating systems and python versions are supported. This can save others from falling for my mistake: I ordered and configuring a server to discover QT does not run on Windows server, some versions of Linux are also not supports -_-'

qubixes (approved):
Approved from my side. @chStaiger perhaps you can take a last look at this, so this can be merged?

.gitignore
```diff
@@ -18,6 +18,7 @@ __pycache__/
 build/
 develop-eggs/
 dist/
+output/
```
chStaiger, line 21: I would still prefer to rename this folder to something like ibridges-build.

build_tools/build_script.py
```diff
@@ -0,0 +1,97 @@
...
+    """Create the exe file for the application.
+
+    args : argparse.Namespace
+        arguments passed to the function
+    """
+    (python, cmd_sep) = uipy.os_specific_settings()
+
+    # Step 1: Setup
+    # Step 1a,(optional) remove the venv to prevent access issues.
+    if args.rem_venv:
```
chStaiger, line 25: The removal of the venv does not seem to work on mac. If I run the command to build the binaries twice I receive:

```
commandline error: /bin/bash: /Users/user/git-repos/iBridges-Gui/venv/bin/activate: Permission denied
```

chStaiger: When I remove the  then the script runs.

chStaiger: Maybe it is handy to have a function called `clear` which removes all of the output of a previous run?

chStaiger (changes requested):
It is really nice to be able to build executables. I verified the build on Mac and all works fine!!
Only when I run the build command twice, there is some error. So I think a "build_clean" function would be nice to add, see my comment in the code.
I also added a small section with instructions for users who want to build an executable to the README.

build_tools/build_script.py
JTvD: Not sure how a `make clean` is going to fix your issue? As the `venv` will be removed in the same way...

Have you tried just placing `exit()` on line 28 to check if the script is allowed to remove the virtual environment? 
I suspect the script is neither allowed to run not remove the virtual environment. 

If that's indeed the case I suggest to leave removing the venv or setting the rights on the folder as a manual step for IOS developers...

.gitignore
JTvD: According to the Nuitka documentation, it can't be a `build `folder. That naming is reserved for the Nuitka build cache.
It is a  `dist` folder, that makes it `ibridgesgui_dist`. 
We probably have to rename it again when making the installer but that is a problem for another day.

README.md
```diff
@@ -64,7 +64,42 @@ The git repository contains a generic *iRODS* graphical user interface.  The iRO
...
+  --code_folder CODE_FOLDER
+                        Full path to the directory with code
+  --ui_folder UI_FOLDER
+                        Full path to the directory with ui files
+  --icons_folder ICONS_FOLDER
+                        Full path to the directory with the icons
+  --venv VENV           Full path to virtual python environment
+```
+
+This will create a virtual environment `venv` in your iBridges folder. Next to that you will find a new folder `build` and `output`. The executable can be found and started like this for Mac and Linux:
```
chStaiger, line 94:
```suggestion
This will create a virtual environment `venv` in your iBridges folder. Next to that you will find a new folder `build` and `ibridgesgui_dist`. The executable can be found and started like this for Mac and Linux:
```

chStaiger (approved):
All good and works! I just did not see that there is a little client to the python script. It is now also documented in the README.

PR #298, opened 2025-01-06
Remove requirements.txt and improve pyproject.toml

qubixes:
We have two ways of installing [optional] dependencies, this merges them into one.


JTvD (approved):
Looks good, this change will also save quite some confusion. 
Only thing I noticed is that flake8 was is the dev requirements file and is not in the toml. 
If this is on purpose it's fine, just wanted to double check.

qubixes:
@JTvD Yes, I removed it on purpose, since it is not used in our continuous integration anymore, and ruff should cover it. Makes sense to harmonize these two.

chStaiger (approved):

PR #301, opened 2025-03-10
300 folder button

chStaiger:
The select_dir in the Browser tab Download button --> Open folders button was not working. Fixed now.


JTvD:
Heey Christine,

There is indeed a small naming difference between pyqt and pyside. Instead of `directory` the term `dir` is used in pyside.
Removing the keyword all together is also fine with me, but I think it was there for a reason? 

Working version:
```           
PySide6.QtWidgets.QFileDialog.getExistingDirectory(
                self, "Select Directory", dir=str(Path("~").expanduser())
            )```

chStaiger:
@JTvD Thanks, I changed the starting level for the destination selection to the home again. If you are fine we can merge.

JTvD (approved):
Looks good to me. The bug is fixed and the functionality is the same as before.

PR #305, opened 2025-03-13
304 search downloadall

chStaiger:
- add download all button for search results
- add label to indicate how may results there are

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The git repository contains a generic *iRODS* graphical user interface.  The iRO
 - Locally from code (for developers)
 
   ```bash
-  git clone git@github.com:chStaiger/iBridges-Gui.git
+  git clone git@github.com:iBridges-for-iRODS/iBridges-Gui.git
   cd iBridges-Gui
   pip install ".[all]"
   ```
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ The git repository contains a generic *iRODS* graphical user interface.  The iRO
 - A specific branch of the git repository (testers, developers)
 
   ```bash
-  pip install git+https://github.com/chStaiger/iBridges-Gui.git@branch-name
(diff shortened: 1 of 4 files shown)


JTvD:
Heey Christine, 

Nice addition, I ran some tests and it works.
However, I do have two small questions on the search tab. 

The first is not directly related to your change but also easy to change, right now I see only 4 rows off results. 
Can we remove some vertical  spacers to create a bit more room for the results on the tab?
I was mainly thinking about the one before 'case sensitive' as that is a part of the metadata filters.
And the space above/below the search button.
spacers

The second one is more on how to implement the selection:
Is it an idea to change 'download selected' and 'download all'  into:
- One button that changes from 'select all' to 'deselect all' that selects the rows in the table.
- A button for the 'Download'

To practical cases I ran into recently:
- I want to download 80% of the data 
- I want to download all but one file
In both cases I would still have to do the selection by hand.

chStaiger:
@JTvD Thanks for the suggestions. I cleaned up a bit and manged to implement the select all button.

JTvD:
@chStaiger nicely done, already increased from 4 rows to 5 rows, does not sound like much but it was an easy win.

There seems to be a bug in the checkbox, maybe it's intentional?  When I do the following:
1. Select a few rows 
2.  Check the box 
Now all the rows, except the already selected ones are selected.

chStaiger:
> @chStaiger nicely done, already increased from 4 rows to 5 rows, does not sound like much but it was an easy win.
> 
> There seems to be a bug in the checkbox, maybe it's intentional? When I do the following:
> 
>     1. Select a few rows
> 
>     2. Check the box
>        Now all the rows, except the already selected ones are selected.

Should be fixed now.

JTvD (approved):
New checkbox works as expected.

chStaiger:
Thanks for your suggestions and testing!!

PR #307, opened 2025-03-18
remove errorlabel and place info box

chStaiger:
Replace errorlabel with info box


JTvD (approved):
Works, also really nice to see it creates each room for another half/complete row in the tables!

PR #310, opened 2025-03-20
Nuitka release pipeline

JTvD:
Add a git workflow for the release which automatically builds the executable. 
A clone of this repository has been used for testing to prevent a lot of test releases.
For trying out the pipeline please use https://github.com/iBridges-for-iRODS/test_releasepipeline

### DIFF
--- a/.github/workflows/nuitka_release.yml
+++ b/.github/workflows/nuitka_release.yml
@@ -1,0 +1,64 @@
+# Build executables on release and upload them as release assets.
+# Based on https://github.com/Nuitka/Nuitka-Action
+
+on:
+  release:
+    types: [created]
+
+jobs:
+ build:
+    strategy:
+      matrix:
+        include:
+          - os: macos-latest
+            tag: macos
+            arch: x64
+          - os: ubuntu-latest
+            tag: ubuntu
+            arch: x64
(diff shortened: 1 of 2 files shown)


chStaiger (commented):
Nice!!! I tested it in the hidden repo.
@qubixes We do not understand why the linux package is much bigger than the other two
https://github.com/iBridges-for-iRODS/test_releasepipeline/releases

Would you have an idea?
Once we figured tat out, I am happy with merging.

qubixes:
> Nice!!! I tested it in the hidden repo. @qubixes We do not understand why the linux package is much bigger than the other two https://github.com/iBridges-for-iRODS/test_releasepipeline/releases
> 
> Would you have an idea? Once we figured tat out, I am happy with merging.

I don't really know to be honest. Looking through the zip files, it looks like there are some extra libraries included in the linux distribution. Maybe for the Mac OS and Windows versions, the ones from the OS itself is used, but it can't be assumed that it is available on linux (since it's more modular). If it's a big problem, then you can make the distribution Mac/Windows only? Most linux users should be able to install it themselves.

qubixes (approved):
Don't really have much to add here, very nice feature!

PR #311, opened 2025-03-21
Update LICENSE

chStaiger:
Now with pyside instead of pyqt we can be a bit more open with our license.


JTvD (commented):
The file change reads very weird. I just have one question, is this the standard LGPL license?

PR #319, opened 2025-05-06
Harmonise cli and gui cached passwords

chStaiger:
The GUI login is extended such that:

- look up all environment files mentioned in the cli config
- if file is not listed or not listed with a password, add to previous gui settings
- only do so if cli config mentions a cached password

In that way the GUI config is leading and the CLI config is used to add, if not defined.

### DIFF
--- a/ibridgesgui/login.py
+++ b/ibridgesgui/login.py
@@ -1,5 +1,6 @@
 """Pop up Widget for Login."""
 
+import json
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


qubixes (changes requested):
I think the integration could be even smoother when the same interface is used. I will test it more thoroughly after that!

qubixes:
@chStaiger It's still not working as expected. It can't find the aliases anymore and neither the passwords from the CLI. Am I missing something?

PR #322, opened 2025-05-15
deprecate ibridges-gui in favour of 'ibridges gui'

chStaiger:
From ibridges 1.5.1 (next bugfix release) one can start the GUI with the ibridges commandline tool by simply typing in

```
ibridges gui
```

This PR just contains the deprecation notice.

### DIFF
--- a/ibridgesgui/__main__.py
+++ b/ibridgesgui/__main__.py
@@ -6,7 +6,6 @@ import os
 import sys
 from functools import partial
 from pathlib import Path
-import warnings
 
 import PySide6.QtGui
 import PySide6.QtUiTools
--- a/ibridgesgui/__main__.py
+++ b/ibridgesgui/__main__.py
@@ -307,10 +307,8 @@ def main():
 def main_deprecated():
     """Deprecate the ibridges-gui call."""
     logger = logging.getLogger(THIS_APPLICATION)
-    logger.info(
-        "'ibridges-gui' will be deprecated. Use 'ibridges gui', needs ibridges 1.5.1 or higher")
-    print(
-        "'ibridges-gui' will be deprecated. Use 'ibridges gui', needs ibridges 1.5.1 or higher")
+    logger.warn(
+        "'ibridges-gui' will be deprecated. Use 'ibridges gui', needs ibridges 1.5.1 or higher.")
 
 if __name__ == "__main__":
     main()
--- a/ibridgesgui/__main__.py
+++ b/ibridgesgui/__main__.py
@@ -307,7 +307,7 @@ def main_deprecated():
     """Deprecate the ibridges-gui call."""
     logger = logging.getLogger(THIS_APPLICATION)
     logger.warning(
-        "'ibridges-gui' will be deprecated. Use 'ibridges gui', needs ibridges 1.5.1 or higher.")
+        "The command 'ibridges-gui' will be deprecated in iBridges 2.0. Use 'ibridges gui' instead.")
 
 if __name__ == "__main__":
     main()
--- a/ibridgesgui/__main__.py
+++ b/ibridgesgui/__main__.py
@@ -302,5 +302,10 @@ def main():
     app.exec()
 
 
-if __name__ == "__main__":
+def main_deprecated():
+    """Deprecate the ibridges-gui call."""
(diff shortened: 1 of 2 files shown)


ibridgesgui/__main__.py
```diff
@@ -302,5 +302,10 @@ def main():
     app.exec()
 
 
-if __name__ == "__main__":
+def main_deprecated():
+    """Deprecate the ibridges-gui call."""
+    print("'ibridges-gui' will be deprecated. Please use 'ibridges gui' from ibridges 1.5.1")
```
qubixes, line 307: You can use a DeprecationWarning, then it will also show up in pytest and the like!

qubixes: Also, personally, I would give people a little bit more time and keep the deprecated command until ibridges 2.0.

ibridgesgui/__main__.py
```diff
@@ -302,5 +302,10 @@ def main():
...
     main()
+
+if __name__ == "__main__":
```
qubixes, line 310: This should call the normal main right?

qubixes (changes requested):
Yes, that's what I had in mind too! Just two comments to mull over.

pyproject.toml
```diff
@@ -53,7 +53,7 @@ all = [
 ]
 
 [project.scripts]
-ibridges-gui = "ibridgesgui.__main__:main"
+ibridges-gui = "ibridgesgui.__main__:main_deprecated"
```
JTvD, line 56: The new entry point should also be listed here right? 
`ibridges gui = "ibridgesgui.__main__:main"`

ibridgesgui/__main__.py
chStaiger: Oh, that is a misunderstanding. I meant the new way of calling the gui is available with ibridges 1.5.1 or higher. I would also like to give people several months to adopt it.

ibridgesgui/__main__.py
chStaiger: Then the main needs to call the deprecation_main ... Is that what you mean?

pyproject.toml
chStaiger: The new "entry point" lies in the ibridges, and no longer in the ibridgesgui package. Since one needs to install ibridges anyways for the gui, we decided to use a command in the ibridges cli which starts the gui.

ibridgesgui/__main__.py
chStaiger: I added it to our normal iBridges GUI log file. Since the tab "log" just shows the log entries as soon as you connect to iRODS the message will not be shown there. It is in the log and it will be printed on the shell.

ibridgesgui/__main__.py
qubixes: The part with `if __name__ == "__main__"` should call the non-deprecated version, not `deprecated_main`.

ibridgesgui/__main__.py
```diff
@@ -302,5 +303,11 @@ def main():
     app.exec()
 
 
+def main_deprecated():
+    """Deprecate the ibridges-gui call."""
+    logger = logging.getLogger(THIS_APPLICATION)
+    logger.warning(
+        "'ibridges-gui' will be deprecated. Use 'ibridges gui', needs ibridges 1.5.1 or higher.")
```
qubixes, line 310: But this is confusing, if you see this message it means you have `1.5.1` or higher, so there's no point?

qubixes, line 310:
```suggestion
        "The command 'ibridges-gui' will be deprecated in iBridges 2.0. Use 'ibridges gui' instead.")
```

chStaiger: It can be that you have ibridgesgui 1.5.1 but still run on ibridges 1.5.0. Unless we also fix that in the pyproject toml. 
Either way would be fine for me. @qubixes Do you have a preference?

chStaiger: Thanks!!! Indeed much better.

JTvD (approved):
Not tested as ibirdges 1.5.1 is not yet released but it looks good!

PR #335, opened 2025-09-29
add last upload and download location to config

### LINKED ISSUE #327: Remember previously chosen download folder in browser download popup screen.
**Is your feature request related to a problem? Please describe.**
I am downloading multiple files from multiple folders into one folder in my windows laptop. I use the browser and then the download button to select and download the files. I do not want to synchronise entire folders, as there are some large files I don't need to download.

**Describe the solution you'd like**
It might speed up the process if in the download popup the windows-folder that I previously chose to download to is given as a default option. I then do not need to click the 'open folders' button anymore for the 2nd to the nth file.

**Describe alternatives you've considered**
Syncin doesn't work as I do not wnat to download all files from the iRODS collection.

**Additional context**

chStaiger:
Add the keywords "last_upload_path" and "last_download_path" to the config. 
When the Upload and Download popup widgets are opened the file selection navigates to those paths rather than to the home. 
If the new config items are not set, the selection will start at the home directory.

### DIFF
--- a/ibridgesgui/config.py
+++ b/ibridgesgui/config.py
@@ -96,12 +96,14 @@ def get_last_ienv_name() -> Union[None, str]:
         return config.get("gui_last_env")
     return None
 
-def get_last_ienv_path() ->  Union[None, str]:
+
+def get_last_ienv_path() -> Union[None, str]:
     """Retrieve the last successfully used environment file."""
     name = get_last_ienv_name()
     if name:
         return name.split(" - ")[1]
     return None
+
 
 def set_last_ienv(ienv: str):
     """Save the last used environment path to the config file.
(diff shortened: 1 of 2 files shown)


JTvD:
Nice new feature, but I am not sure if this is the desired functionality. 
The issue explicitly mentions: "I then do not need to click the 'open folders' button anymore for the 2nd to the nth file."
In the current implementation the user still needs to press this button before the last download path is retrieved.

The upload works in a similar way, using the iRODS path currently open in the browser as the target location.

chStaiger:
Indeed, I am still thinking of integrating a multi select of iRODS items. But I am not sure if the browser biew is the right place for it. I am considering creating an extra Download view which is not loaded by default but choosable from the menu "Views" for such tasks. Then I can also integrate the Data owner selection criterion and timestamps.

PR #336, opened 2025-09-29
GUI can be started from ibridges session directly

### LINKED ISSUE #318: Integrate better with the iBridges CLI
It would be nice if the GUI and CLI would integrate better. Now, with the new shell PR, this should be easier and more reliable. What I would like to see:

- Aliases in the CLI should show up in the GUI.
- The currently selected environment should be immediately connected to if possible. Disconnecting and reconnecting to another server should be possible still of course.
- `ibridges-gui` should become `ibridges gui`. This should be builtin to the default subcommands. If the gui is not installed, instructions should be shown on how to install it.

To reduce the need for maintenance, the gui and cli should also use the same functions and classes to do similar things (like aliases). So integrate with:

- Authentication: `ibridges.cli.util.cli_authenticate`
- Configuration editing: `ibridges.cli.config.IbridgesConf`

chStaiger:
One can now skip the login and start the gui wioth a session directly:

```
from ibridges.interactive import interactive_auth
from ibridgesgui.__main__ import main

session = interactive_auth()
main(session)
```

### DIFF
--- a/ibridgesgui/__main__.py
+++ b/ibridgesgui/__main__.py
@@ -12,7 +12,6 @@ import PySide6.QtGui
 import PySide6.QtUiTools
 import PySide6.QtWidgets
 import setproctitle
-
 from ibridges import Session
 
 from ibridgesgui.browser import Browser
--- a/ibridgesgui/__main__.py
+++ b/ibridgesgui/__main__.py
@@ -6,11 +6,14 @@ import os
 import sys
 from functools import partial
 from pathlib import Path
+from typing import Optional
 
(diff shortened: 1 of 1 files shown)


chStaiger:
@qubixes Could you please test that and integrate it into the CLI shell?

qubixes:
@chStaiger The gui command now works in the shell, but unfortunately the shutdown of the gui is not compatible. Maybe it just does something like sys.exit()? If that is fixed, it should work.

chStaiger:
I think I solved it...

PR #341, opened 2025-11-05
low res compat

### LINKED ISSUE #330: minimum size on small screens
As part of some pre-release testing we installed ibridges on various machines and noticed that while it works fine the  minimum size is too big for the 13" laptops we have:
Image

After lowering the minimum size of the mainmenu.py from 1300 by 850 to 1000 by 700 it looks like this.
Not perfect but at least everything fits on the screen now:
Image

Can we set more widgets to scale to make the GUI available on these devices as well?
(We can leave the preferred size the same so nothing changes for normal users besides that the application allowes you to make the screen smaller)

maartenschermer:
Fixes #330 

When on starting the available screen width (taken from `app.primaryScreen().availableGeometry()`) is <=1280 px, the app uses alternative .ui files for the main screen and various tabs (`MainMenu_small.ui`, `tabBrowser_small.ui`, etc). These contain the same elements as their full size counterparts, but with a slightly rearranged lay-out, less spacers, and smaller font size to make everything fit on the screen comfortably.

I did _not_ manage to get the application to start with a maximized screen. I assume it's possible, but exactly how currently eludes me. Instead, the user should double-click the program's top bar, click the 'maximize window'-button, or press F11 (see below) to toggle between normal and maximized window.

Additionally:
+ The cursor turns into a "working"-style cursor from the moment user clicks 'Connect', until the browser tab appears (or connecting fails).
+ The items in the local filesystem tree in the Sync-tab are now alphabetically sorted.
+ The local filesystem tree was designed to use the user's home folder (`PySide6.QtCore.QStandardPaths.StandardLocation.HomeLocation`) as its base, but it didn't quite work, and it ended up using the client's root-folder instead. I've now properly implemented the client's root-folder as base by design (using `os.path.abspath(os.sep)`; should be platform independent). If it should instead use the home location as base, I can change that (note: the user cannot navigate out of the base folder).
+ I've added a few extra filetypes to the preview-function of remote files (`["txt", "json", "csv", "md", "yaml", "R", "py"]`).
+ Pressing F11 toggles between normal and maximized window.


JTvD:
@maartenschermer, nice work. I had hoped you could get away with automatically resizing UI elements. Implementation wise this is a simple alternative. On my first test I noticed a few things. This might also be because I had to go below the 1280 to trigger the small UI's:

- The logo does not scale yet, should be a small one: 
logo

- The browser tab is close but not yet perfectly aligned:
- resize_browser

- The up/download buttons are not aligned, could also be an image sizing issue?
resize_sync

So far I only did a visual check on a small screen, I have not yet tried the buttons.

maartenschermer:
Thanks for the feedback. I've fixed the logo, but the other ones are trickier. I've made them to fit 1280x800 (see images) - any lower will probably won't fit as well. I've found it's not that easy to make all elements scale proportionally. I'm not sure what resolution you tried, but is it feasible anyone will need such a low resolution?

Screenshot from 2025-11-07 16-19-41
Screenshot from 2025-11-07 16-20-35

JTvD:
@maartenschermer when running on 1280 pixels it looks a lot better.
(Note, I changed the line mentioned above to get it to load the small UI)
When running the application full screen the logo's look good. There is however one small detail,  by default the UI does not open full screen. It opens as small windows just like the normal sized UI: 
default opening size

As you mention scaling down to this size is tricky, you could opt to set it full screen by default in the low resolution mode. 
Maybe also hide the minimize button?

PR #342, opened 2025-11-06
upload metadata

### LINKED ISSUE #340: Add metadata to upload

chStaiger:
- Adding json schema for ibridges metadata
- Extensing the upload with metadata uploads

@qubixes Can we move the metadata json and the validation function to the iBridges package?


qubixes:
@chStaiger Yes, we could. I'm not sure how happy I am with the interface anyway, so it might be a good time to do a little bit of rethinking.

ibridgesgui/gui_utils.py
```diff
@@ -97,6 +101,48 @@ def populate_textfield(text_widget, text_by_row: Union[str, list]):
...
+        raise ValueError(f"Metadata file not found: {md_path}")
+
+    try:
+        with md_path.open("r", encoding="utf-8") as f:
+            data = json.load(f)
+    except json.JSONDecodeError as e:
+        raise ValueError(f"Invalid JSON in metadata file: {e}") from e
+
+    # Load the schema from the package
+    with (
```
qubixes: Split in two?

ibridgesgui/gui_utils.py
```diff
@@ -97,6 +101,48 @@ def populate_textfield(text_widget, text_by_row: Union[str, list]):
...
+        pkg_resources.files(ibridgesgui.md_schemas)
+        .joinpath("ibridges_metadata_schema.json")
+        .open("r", encoding="utf-8") as f
+    ):
+        schema_data = json.load(f)
+
+    try:
+        validate(instance=data, schema=schema_data)
+    except ValidationError as e:
+        raise ValueError(f"Metadata validation failed: {e.message}") from e
```
qubixes: Why convert a ValidationError in a ValueError?

ibridgesgui/gui_utils.py
```diff
@@ -142,6 +188,7 @@ def combine_operations(operations: list[Operations]) -> Operations:
     ops = operations[0]
     ops.create_dir = set().union(*[o.create_dir for o in operations])
     ops.create_collection = set().union(*[o.create_collection for o in operations])
+    ops.meta_upload = set().union(*[o.meta_upload for o in operations])
```
qubixes, line 203: I don't understand why this seems so complicated. Isn't it just making a copy?

ibridgesgui/popup_widgets.py
```diff
@@ -377,15 +501,15 @@ def _start_upload(self, lpaths):
             ops = combine_operations(
                 [
                     upload(
-                        p,
+                        path,
                         self.irods_path,
                         overwrite=self.overwrite.isChecked(),
                         dry_run=True,
+                        **({"metadata": metadata} if metadata is not None else {})
```
qubixes: Why doesn't metadata=metadata not work?

qubixes (changes requested):
I have made a quick scan of the code and have a few points of attention. Otherwise, I haven't checked whether the code works correctly (yet).

ibridgesgui/gui_utils.py
```diff
@@ -13,7 +15,9 @@
 import PySide6.QtWidgets
 from ibridges import IrodsPath
 from ibridges.executor import Operations
+from jsonschema import ValidationError, validate
```
JTvD, line 18: not a default python library, I had to add it to the toml file

ibridgesgui/ui_files/uploadData.py
```diff
@@ -1,30 +1,19 @@
-# -*- coding: utf-8 -*-
+# Form implementation generated from reading ui file 'ibridgesgui/ui_files/uploadData.ui'
+#
+# Created by: PyQt6 UI code generator 6.10.0
```
JTvD: File was made with pyqt6, recreated it locally with pyside6 to test

ibridgesgui/ui_files/uploadData.ui
```diff
@@ -149,26 +100,54 @@ QTabWidget#info_tabs
...
-        <string>Upload</string>
+        <string>Select Folders</string>
        </property>
       </widget>
      </item>
      <item>
-      <spacer name="horizontalSpacer_2">
+      <widget class="QPushButton" name="file_button">
+       <property name="text">
+        <string>Select Files</string>
```
JTvD, line 145: This text suggest that you can select multiple files, while the underlaying code allows only one file at the time.  
for files this can easily be changes by using the multi file selector instead of a single file selector, which returns a list changing the function to: (cntrl+click the files)
```
    def select_file(self):
        """Open file selector."""
        last_upload_path = config_get_last_upload_path()
        if not last_upload_path:
            last_upload_path = Path("~").expanduser()
        select_files = PySide6.QtWidgets.QFileDialog.getOpenFileNames(
            self, "Open File", dir=str(last_upload_path)
        )
        for selected_file in select_files[0]:
            path = self._fs_select(selected_file)
            if path is None or str(path) == "." or path in self.get_all_paths():
                continue
            config_set_last_upload_path(Path(path).parent)
            self.add_row(path, "")
```

ibridgesgui/ui_files/uploadData.ui
```diff
@@ -149,26 +100,54 @@ QTabWidget#info_tabs
...
+    </widget>
+   </item>
+   <item row="0" column="2">
+    <layout class="QHBoxLayout" name="horizontalLayout_3">
      <item>
-      <widget class="QPushButton" name="upload_button">
+      <widget class="QPushButton" name="folder_button">
        <property name="text">
-        <string>Upload</string>
+        <string>Select Folders</string>
```
JTvD: Pyside does not have a default way to select multiple folders, we can create this if you want? But you could also change the text to 'select folders' or we can use a normal QFileDialog to allow the selection of both files and folders

ibridgesgui/md_schemas/ibridges_metadata_schema.json
```diff
@@ -0,0 +1,54 @@
+{
```
JTvD, line 1: Don't forget to update the build_script.py so it copies the md_schemas

JTvD (changes requested):
The metadata upload in the browser upload is a nice addition. What I mis is a way to add metadata after the upload completes. For example in the case the file uploads succeeded but the metadata upload failed. 

After the changes I would like to play a bit more with uploading different metadata files.

ibridgesgui/ui_files/uploadData.py
chStaiger: Thanks for spotting this. Will be corrected.

ibridgesgui/gui_utils.py
chStaiger: Thanks. I added it!

ibridgesgui/gui_utils.py
chStaiger: You mean separate the loading of the file from the actual validation?!

ibridgesgui/gui_utils.py
chStaiger: Yeah the ValueError is one level further down. Will turn it into a a ValidationError.

ibridgesgui/gui_utils.py
chStaiger: The input are several ops objects in a list. I take the first one and then update this with the operations from the other ops objects. I need the set, since single operations can potentially be listed several times.

chStaiger:
> The metadata upload in the browser upload is a nice addition. What I mis is a way to add metadata after the upload completes. For example in the case the file uploads succeeded but the metadata upload failed.
> 


I would put this into the metadata tab. So you would navigate to the collection to which you want to upload, the browser is refreshed after the upload and then you select the collection or file you want to decorate with metadata from the file and you would use the metadata tab for that. Would that work?

ibridgesgui/ui_files/uploadData.ui
chStaiger: Thanks! That is a nice addition.

ibridgesgui/popup_widgets.py
chStaiger: It works ... but this look much cooler :P.
Thanks for the hint, I adjusted the code.

ibridgesgui/ui_files/uploadData.ui
chStaiger: I adjusted the button label

ibridgesgui/md_schemas/ibridges_metadata_schema.json
chStaiger: Done! Tested it on Mac. The medata schemas are in the main distribution folder (same level as icons). I also created a switch in the code for reading them in since that differs from the python packge.

JTvD:
> I would put this into the metadata tab. So you would navigate to the collection to which you want to upload, the browser is refreshed after the upload and then you select the collection or file you want to decorate with metadata from the file and you would use the metadata tab for that. Would that work?

Interesting idea, I can see multiple reasons to create a metadata tab. 
Also to create/edit schema's.
However, this functionality requires a file browser to select a data object/collection to add metadata to.
I am not sure if that's wise or weather it would be easier to add a button to add metadata files in the browser tab. 
Something to think about.

chStaiger:
> > I would put this into the metadata tab. So you would navigate to the collection to which you want to upload, the browser is refreshed after the upload and then you select the collection or file you want to decorate with metadata from the file and you would use the metadata tab for that. Would that work?
> 
> Interesting idea, I can see multiple reasons to create a metadata tab. Also to create/edit schema's. However, this functionality requires a file browser to select a data object/collection to add metadata to. I am not sure if that's wise or weather it would be easier to add a button to add metadata files in the browser tab. Something to think about.

I did not mean an own Metadata tab ;) For now I would add a new button to the metadata tab in the Browser view which shows all metadata whern you select an item from the table.

PR #343, opened 2025-12-02
allow to filter for collections and objects

chStaiger:


docs/faq.qmd
```diff
@@ -0,0 +1,34 @@
...
+
+You can omit that only by using the [iBridges API](https://ibridges.readthedocs.io/en/latest/faq.html#my-metadata-with-keys-starting-with-org-do-not-show-up-in-the-ibridges-metadata).
+
+## How can I download files that follow a certain pattern but do not lie in the same collection?
+
+You can use the *Search* for this. With the search mask you can filter for specific data no matter where they are located. In the results table you can select those files you would like to download.
+
+## How do I find all the data I have access to?
+
+Shared data objects always needs to lie in a collection that you can access, it's an iRODS feature.
```
chStaiger:
```suggestion
When data is shared in iRODS it is not always easy to know where the shared data lies as they are not visible from your "home" collection or project collection.

However, shared data objects always need to lie in a collection that you can access, it's an iRODS feature.
```

maartenschermer:
Looks good, good addition. Bit odd that selecting both collection and objects gives the same reults as deselecting them both. Maybe a QRadioButton with three options (collections, objects, both [default]) would be more fitting?

PR #344, opened 2025-12-04
fill download loc and bigfixes

chStaiger:
Implements #327 for Downloads.

The download location last used will be stored in the config. When hitting the download button in the browser view, the last location will be checked and loaded if it exists.


JTvD (approved):
Nice work! Tested a few scenario's which all worked as expected:
- Normal download to same folder twice
- Download to deleted folder
- Restart program and try again

PR #346, opened 2026-01-29
add context column to resc table

### LINKED ISSUE #345: resource context in info tab
On the server we have context set on the resources. 
These are returned by the call:
https://github.com/iBridges-for-iRODS/iBridges-GUI/blob/22eb0a311b428a259bea4d1db8cf6a628d1d3cf6/ibridgesgui/info.py#L48

Image

However, the resources table only has 3 columns:
name, status, free space

So when it tries to set the values from the context a new row is created in the table:

Image

chStaiger:
The default number of rows is 4 since pyside fills up the empty space that the widget has with them. I added a fourth column.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,8 @@ Please note, that this feature is still in experimental state.
 
 ### Prebuilt executables
 Go to the [Releases page](https://github.com/iBridges-for-iRODS/iBridges-GUI/releases). In the latest releases you will find three zip-files, these contain the prebuilt application.
-Download the respective zip-file for your system and unpack it. In the unpacked folder there is a file `ibridges_gui.[exe, sh]`
-For Windows simply click on the exe-file, for Mac and Linux open the file with a shell `bash ./ibridges_gui.sh`.
+Download the respective zip-file for your system and unpack it. In the unpacked folder there is a file `ibridges_gui.[exe, bin]`
+For Windows simply click on the exe-file, for Mac and Linux open the file with a shell `bash ./ibridges_gui.bin`.
 
 
 ### Building Executables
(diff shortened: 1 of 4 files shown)


chStaiger:
We could do something about the "None"s in the table. I do not dare to fill them with blank lines, since our SURF instance has a context with '   '. So there is  difference in information between blanks and None. If you have an idea how to translate None into something more userfriendly, let me know.

JTvD:
> The default number of rows is 4 since pyside fills up the empty space that the widget has with them. I added a fourth column.

Just ran some tests, if you look closely at: https://github.com/iBridges-for-iRODS/iBridges-GUI/blob/22eb0a311b428a259bea4d1db8cf6a628d1d3cf6/ibridgesgui/info.py#L49
you see that `populate_ table` is called with `len(resc_info[0])`  which is always 4 as the items for each resource: the Name, Status, Free space and Context. 

Updating the line to:
`populate_table(self.resc_table, len(resc_info), resc_info)`
removes the empty line.

JTvD:
> We could do something about the "None"s in the table. I do not dare to fill them with blank lines, since our SURF instance has a context with ' '. So there is difference in information between blanks and None. If you have an idea how to translate None into something more userfriendly, let me know.

Good point, is it an idea to check is the status is `None` here:
https://github.com/iBridges-for-iRODS/iBridges/blob/b22cf0add1706671787e00243b1d922e6ff1cecf/ibridges/resources.py#L136
and convert it to `-` or indeed an empty string? 

If you do it in iBridges and only for the status field I think you avoid the issue you mention about accidentally converting a context. What do you think?

chStaiger:
Thanks!! You are right. It is fixed.

JTvD (approved):
Looks good, one bug down.

PR #348, opened 2026-02-26
Refactor browser and add tests

chStaiger:
Hi all, 

I started refactoring the GUI. My main concern is, that we want to have tests and I also figured that there is still a bit of clutter between the GUI logic and the iRODS logic.

This is my first attempt for one of the most complex tabs. Please let me know what you think of the idea of refactoring and also if you have any preferences.

I hope you like my first ideas.


ibridgesgui/irods_browser_service.py
```diff
@@ -0,0 +1,143 @@
+"""iRODS functionality fro browser."""
```
chStaiger:
```suggestion
"""iRODS functionality for browser."""
```

tests/browser/test_browser_controller.py
```diff
@@ -0,0 +1,140 @@
...
+    path = MagicMock()
+    path.collection_exists.return_value = True
+    controller._item_path = MagicMock(return_value=path)
+    controller._set_path = MagicMock()
+
+    controller._open_selected_path()
+    controller._set_path.assert_called_once_with(path)
+
+
+def test_delete_data_success(controller, ui):
```
JTvD: Is there a reason not to test the other flow? 
I would be really sad if I press 'no' in the pop up and the file is still deleted...

You can also argue about testing the exceptions but it's way less problematic if they contain a bug

JTvD:
To make it easier to check if the tests are checking all the lines of code I suggest to sett up [coverage](https://pypi.org/project/coverage/ ) or pytest-cov first.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: |
+          echo "QT_QPA_PLATFORM=offscreen" >> $GITHUB_ENV
+          echo "QT_MAC_WANTS_LAYER=1" >> $GITHUB_ENV
+
+      # -----------------------------
+      # Run tests 
+      # -----------------------------
+      - name: Run tests and coverage
+        run: |
+            python -m pytest --cov=ibridgesgui --cov-report=term-missing tests
```
qubixes, line 72: You might want to set a lower limit on which percentage of coverage you require.

qubixes (commented):
I don't feel like I'm qualified enough to really have a strong opinion on the new structure. It does seem to adhere to the MCV design pattern, which I seem to remember from QT documentation.

.github/workflows/tests.yml
chStaiger: Definitely! For now I wanted to have a very detailed report for the time period in which I transfer all of the classes to the new pattern. I would also still need to identify the classes that require these meticulous tests and exclude the other ones. Once the code is refactored I planned to set the mark at at least 80%.

maartenschermer:
Looks good (and works as it should, as far as I can tell), though the PR is too big to go through in detail.

chStaiger:
@qubixes approved through personal communication.

PR #354, opened 2026-04-22
Refactor threads

chStaiger:


tests/threads/test_sync_thread.py
qubixes, line 1: Is this module name correct?

qubixes (approved):
Not much to comment on, good to have more tests!

PR #375, opened 2026-05-28
Small fixes

chStaiger:
- set cursor arrow back after upload
- Fix label in resource table
- Fix label in ACL table: Owner -> Creator
- Improve preview: sniff mime type and show in preview if it is text
- Filtering out metadata ACLs https://github.com/iBridges-for-iRODS/iBridges/issues/412
- Fix metadata update: reduced to meta.add and then meta.delete

### DIFF
--- a/ibridgesgui/browsertab/irods_browser_service.py
+++ b/ibridgesgui/browsertab/irods_browser_service.py
@@ -2,11 +2,11 @@
 
 import logging
 from typing import Iterable, Tuple
-from irods.exception import CAT_NO_ACCESS_PERMISSION
 
 from ibridges import IrodsPath
 from ibridges.permissions import Permissions
 from ibridges.util import obj_replicas
+from irods.exception import CAT_NO_ACCESS_PERMISSION
 
 from ibridgesgui.gui_utils import get_irods_item
 
@@ -47,10 +47,11 @@ class IrodsBrowserService:
     # -------- preview ---------
 
     def is_text_bytes(self, data: bytes) -> bool:
+        """Sniff if file contains text."""
         # Null bytes → definitely binary
         if b"\x00" in data:
             return False
-    
+
         # Try common encodings
         for enc in ("utf-8", "latin-1", "cp1252"):
             try:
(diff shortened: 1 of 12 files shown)


ibridgesgui/browsertab/irods_browser_service.py
```diff
@@ -134,20 +176,53 @@ def update_metadata(
...
+                raise irods_exc
+            raise err
 
     # -------- ACLs / permissions --------
 
+    def get_acl_strings(self):
+        PERM_MAP = {
+            "read_object": "read",
+            "modify_object": "write",
+            "null": "delete",
```
trel: no own?

chStaiger: `own` stays own from the prc:

```
In [6]: path = IrodsPath(session, "/tempZone/home/irods1")
In [7]: perm = Permissions(session, path.collection)
In [8]: perm.available_permissions # prc function


OrderedDict([('null', 1000),
             ('read_metadata', 1040),
             ('read_object', 1050),
             ('create_metadata', 1070),
             ('modify_metadata', 1080),
             ('delete_metadata', 1090),
             ('create_object', 1110),
             ('modify_object', 1120),
             ('delete_object', 1130),
             ('own', 1200)])
```
The `PERM_MAP` maps some acl mode names to something less ambiguous. As you see the available permission modes for a collection are `*_object` while in the above code I am looking at a collection.

So with 

```
avail = [PERM_MAP.get(perm_str, perm_str) for (perm_str, _) in perm.available_permissions.items()]
``` 
I am only changing the strings how they appear in the GUI.

chStaiger: I must admit I am a bit confused with the differences between 

```
             ('read_object', 1050),
             ('create_object', 1110),
             ('modify_object', 1120),
             ('delete_object', 1130),
```

### Data objects
- `create_object`: "read" + "create_metadata"
- `modify_object`: no delete, all other actions on file and metadata work
- `delete_object`: delete and overwrite (put -f)

### Collections (tested wit icommands)
- `create_object`:  returns CAT_NO_ACCESS_PERMISSION when trying to upload a file
- `modify object`: create data objects and collections, add metadata, mod metadata, rm metadata
- `delete_object`: I can also upload data to the collection.

trel: > create_object: returns CAT_NO_ACCESS_PERMISSION when trying to upload a file
> modify object: create data objects and collections, add metadata, mod metadata, rm metadata
> delete_object: I can also upload data to the collection.

the 'object' here is a collection.
when you think of it that way... the findings you have make more sense I think...

create object - means you can create a collection itself
modify object - means you can edit things about that collection, including adding files / children objects/collections
delete object - you can delete a collection

PR #378, opened 2026-06-30
Expose ibridges-gui entrypoint in pyproject.toml

dometto:
This is so users can `uv tool install ibridgesgui`. This (or `pipx install` if you don't want `uv`) is current best practice for installing executable python apps: `uv tool` ensures the app is isolated in its own virtual environment and the entrypoints become available on your PATH (no matter what venv you are in). You can also use `uv tool` to install an application for all users on the system.

You can test this already with `uv tool install git+https://github.com/dometto/iBridges-GUI.git@add_entrypoint`

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,19 @@ You can find the latest release [here](https://github.com/iBridges-for-iRODS/iBr
 The build was done using the latest version on an x64 architecture.
 
 ## Installation
-- The python package 
+- Install the python package into its own virtual environment (recommended):
+
+  ```bash
+  uv tool install ibridgesgui # installs for your user by default; see uv tool options for how to install for all users on the system
+  ```
+
+  OR
+
+  ```bash
+  pipx install ibridgesgui
+  ```
+
+- Install the python package in the current virtual environment (or system-wide):
 
   ```bash
   pip install ibridgesgui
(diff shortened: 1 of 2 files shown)


dometto:
I've called the entrypoint `ibridges-gui` in line with [this](https://github.com/iBridges-for-iRODS/iBridges-GUI/blob/develop/ibridgesgui/__main__.py#L14)

It's slightly annoying that you have to install `ibridgesgui` and run `ibridges-gui`. We could a) change the entrypoint to `ibridgesgui` b) add a second entrypoint `ibridgesgui`. Or we leave it as is :)

chStaiger:
We deprecated the command `ibridges-gui` in v2. We are using now the shell to call the gui: `ibridges gui`. So the entrypoint does not match anylonger. The `ibridgesgui` package is depending on `ibridges` which installs also the CLI. Would that work for you?

dometto:
> Would that work for you?

Nope :-) I would like to be able to use `uv` to install ibridges-gui as a tool for all users on a system (on SRC).The normal way to do that is via `pipx` or `uv tool install`. What is the rationale for making this impossible? (Note that we cannot use a global pip install, and creating a venv and exposing the executables in it manually is not very userfriendly).

dometto:
Hmm, just noticed I can do `uv tool install ibridges --with ibridgesgui` and then `ibridges gui`. That kind of works!

Still, what's the reason for not providing an entrypoint for this package?

(Fine to close this if there's a good reason, but I would consider adding `uv tool install ibridges --with ibridgesgui` to the Readme)

chStaiger:
The GUI and the CLI were moved closer together, drawing on the same configuration files and cached information, hence it was logical to use also the CLI to start the GUI. In that sense the GUI is seen as a real extension of the API and the CLI.

chStaiger:
Thanks for figuring out how to install the tools system wide. That is something we will test next time before we create a new release.

